Issue #25, opened 2020-07-30
Comments on "core" profile and scope of constraints


mjkoster:
From the July 30 discussion, it seems like the word "core" is indeed subject to much interpretation, and we need to re-orient the discussion on the scope of constraints.

I think the best definition of what is desired in "core" is to enable out-of-the box interoperability for "consumers" meaning as I understand API consumers rather than someone shopping at Best Buy, but maybe that also. 

Who are the consumers?

We didn't discuss as much who the profile is for, what class of users, what application domain. Saying we want OOB interoperability across all application domains covered by Thing Description seems difficult. Defining narrow constraints that enable OOB interoperability for constrained devices seems to me to be a much narrower scope than we have for W3C WoT in general, and should be clearly defined as one profile that is targeted for a specific use case. 

**local/vertical scope** - what does this profile constrain, and what is allowed to be adapted to in the endpoints? A profile for digital twin interaction might be optimized for purpose by allowing multiple protocols, e.g. HTTP and MQTT and favor adaptation. A profile for battery powered smart home/building devices would like CoAP or CHIP and can't be expected to adapt much.

**horizontal scope** - which application domains does the profile optimize for? Are we calling "the web" a domain at the same time we're expecting to support constrained devices that can't adapt to a few base protocols? Should we instead target profiles to stakeholder groups where we enable smart home/building as an ecosystem, etc.?

What if we want to create a _WoT Profile_ to enable TD to be used to automate industrial fieldbus deployments and add digital twin support through OMA LWM2M objects and data models that are compatible IEC61499 function blocks. We might need to define some network interaction model and "protocol bindings" for some fieldbuses like OPC-UA, modbus, and BACnet. The out of box interoperability scenario would be different from consumer smarthome but still quite important.

If we create a profile that is required, would it apply to both smart home/building and industrial fieldbus profiles? What is the tradeoff between narrow scope for the experience of some users vs. broader scope for coverage across industry? If we apply one "core" profile to all of WoT, is that consistent with our overall objectives?

benfrancis:
The way I think about this is to imagine a smartphone app which acts as a generic Web of Things browser/client (unfortunately as far as I know this doesn't exist yet, but I'd love to create one). What are the constraints that the developer of a greenfield device (or gateway/cloud service adapting a brownfield devices) must follow in order to guarantee out-of-the-box interoperability with that client?

[...]

Issue #21, opened 2020-07-30
Term 'Core' may cause a false impression 


sebastiankb:
As I mention in last week web meeting we have to be very careful about wording. I'm not sure if the term 'core' or 'Core Profile' results to the assumption that this core have to be always be implemented (since it is a 'core') even I would actually follow a profile that covers CoAP or MQTT.

A proposal would be that we use the term 'default' as alternative. 'default' has the impression that there are some basic settings / restrictions that will be adopted if they are not changed otherwise.

benfrancis:
I agree with this, "default" may be a better name.

danielpeintner:
I agree that "default" is better than "core" but I raised the question in https://github.com/w3c/wot-profile/pull/20#discussion_r462826017 whether it makes sense to name it according to the technology used (e.g., "http-json")

benfrancis:
> I agree that "default" is better than "core" but I raised the question in #20 (comment) whether it makes sense to name it according to the technology used (e.g., "http-json")

+1 That sounds like a great idea.

sebastiankb:
An "http-json" can be also an option that gives a clear technology context. However, I would be worried if this results to bulk of profiles in the future with "http-yaml" or  "http-xml" or "http-cbor" or "coap-json" etc. 

Btw: I also like the term "OOB Profile". I have seen this here #25.

egekorkan:
I don't remember where I had mentioned it before but I think that instead of defining a profile, a set of features supported by a TD could be communicated to the Consumers. From this discussion, it seems that people might be more interested in such an "annotation". So instead having something like:

```json
{
    "@context": "https://www.w3.org/2019/wot/td/v1",
    "title": "MyLampThing",
    "profile":"http-json",
    "securityDefinitions": {
        "basic_sc": {"scheme": "basic", "in":"header"}
    },
    "security": ["basic_sc"],
    "properties": {
        "status" : {
            "type": "string",
            "forms": [{"href": "https://mylamp.example.com/status"}]
        }
    }
}
```

We can have something like:

```jsonc
{
    "@context": "https://www.w3.org/2019/wot/td/v1",
    "title": "MyLampThing",
    "features":["http","json"], //Only difference here
    "securityDefinitions": {
        "basic_sc": {"scheme": "basic", "in":"header"}
    },
    "security": ["basic_sc"],
    "properties": {
        "status" : {
            "type": "string",
            "forms": [{"href": "https://mylamp.example.com/status"}]
        }
    }
}
```

[...]

PR #47, opened 2020-10-07
Issue 45 - FPWD clarifications on profile name and figures

mlagally:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/47.html" title="Last updated on Oct 7, 2020, 11:23 AM UTC (d5e6f7f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/47/2b8cf9f...d5e6f7f.html" title="Last updated on Oct 7, 2020, 11:23 AM UTC (d5e6f7f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -136,10 +136,10 @@
       of the core profile, however the interoperability guarantees of the core profile
       hold only for the <a>WoT Core Profile</a> subset.
 
-      <section class="ednote">
-        The name <a>WoT Core Profile</a> is still under discussion in the group and is used as a working title.
-        It is subject to change after the profile specification has reached a certain level of maturity.
-      </section>
+    <section class="ednote">
+      The name <a>WoT Core Profile</a> is still under discussion in the group and is used as a working title.
+      It is subject to change after the profile specification has reached a certain level of maturity.
+    </section>
 
     <section>
       <h3>Motivation for a Profile</h3>
@@ -554,6 +554,10 @@
         to multiple profiles. Profiles can build on top of each
         other or overlap, extended profiles can be built on top
         of the core profile.</p>
+      <p>
+        This specification does not put any requirements on the 
+        scope and contents of other profiles.
+      </p>
 
       <figure id="WoT Profiles">
         <img src="images/WoT Profiles.png" class="wot-profiles" alt="WoT Profiles" />
@@ -848,7 +852,7 @@
             lamp1_color_g and lamp1_color_b. A similar
             mapping can be done for arrays and hierarchical
             objects. This constraint leads to simpler Thing
-              Descriptions that can be handled by very
+            Descriptions that can be handled by very
             limited devices.
           </p>
         </section>
--- a/index.html
+++ b/index.html
@@ -136,6 +136,11 @@
       of the core profile, however the interoperability guarantees of the core profile
       hold only for the <a>WoT Core Profile</a> subset.
 
+      <section class="ednote">
+        The name <a>WoT Core Profile</a> is still under discussion in the group and is used as a working title.
+        It is subject to change after the profile specification has reached a certain level of maturity.
+      </section>
+
     <section>
       <h3>Motivation for a Profile</h3>
     </section>


mmccool:
So yet again, I want to say that I agree with the "core" name if it is in fact defined to be the set of constraints common to all profiles (eg http-specific profiles, etc).

mmccool:
So I had some further thoughts on this:
1. I would like to confirm whether Sebastian has the "target specific verticals" use case in mind for profiles.  I think this is a different use case than "out of the box interoperability".  We need to resolve whether or not we want to extend the scope of the profile spec to support profiles for "verticals" as well (which may indeed have other constraints than what we currently have in the "core").
2. Assuming we resolve to the scope extension in 1, then I would propose we have two "categories" of profiles: "core" and "non-core" (or maybe "vertical").  The former's primary goal is interoperability.  If you as a user want that, then you select it.  If you need to work with pre-existing brownfield systems in particular verticals, however, you might have to select a non-core profile aimed at those verticals.
3. We could rename "core" to "interop".

egekorkan:
Why not _simply_ name the current core profile to _simple_ profile? This is a profile that is easy to implement, uses very conventional mechanisms and protocols, and would achieve interoperability by being easy, not necessarily because many people have already implemented Things that respect this. If this profile is created because it was identified that the Thing implementations by the WG respect this profile _already_, this claim should be proven. We have all the data to show whether this is the case or not.

I also strongly object naming any profile core since, in an environment like Web that is constantly changing, no set of technologies can be core for more than 5 (10?) years. I have also mentioned somewhere that we can name profiles by more atomic features, like `"profile":["http","json","sse","simpleDataSchema", "humanMetaData"]`. Then, we can say that if all these atomic profiles are respected, it can be called _myProfile_. I also don't like _interop_ since it would give the impression that other profiles do not provide interoperability.

danielpeintner:
Note: there is a dedicated issue w.r.t. the name, see https://github.com/w3c/wot-profile/issues/21
I suggest having discussions there ;-)

mlagally:
We are mixing two topics in this conversation, the profile **name** and the profile **scope and use cases**.
Let's treat them separately to make sure we address the individual qustions.

We already did a resolution on the **scope** of the profile in several arch calls back in July:

> RESOLUTION: Accepting the following requirements for the FPWD: Interoperability, Limit and reduce complexity, Eliminate ambiguities, Limit resource consumption, finite set of features and capabilities, follow security and privacy best practices subject to group feedback.

The current FPWD draft addresses this scope.

**Out of the box interoperability** is the main objective of the core profile. The conformance section provides the corresponding guarantees. 

@egekorkan Please consider that deployed WoT devices are different to Web devices, i.e. in many cases they don't undergo regular SW maintenance and updates. There's no business model for providing updates, which is the reason why the security problems we see in many IoT devices (e.g. IP cameras) don't get fixed.
Nothing is set in stone in the Web and in WoT forever. If we define a Core 1.0 which will hold for 5 years we whould have done an excellent job. There's always room for defining other profiles or a new version of the FPWD profile, if there are new requirements.

sebastiankb:
@mlagally 
>We are mixing two topics in this conversation, the profile name and the profile scope and use cases.
Let's treat them separately to make sure we address the individual qustions.

+1

I think, for the name issue we already clarified this yesterday and the corresponding editorial note is included. Regarding the name I already shared my view in July on this:

https://github.com/w3c/wot-profile/issues/21

>RESOLUTION: Accepting the following requirements for the FPWD: Interoperability, Limit and reduce complexity, Eliminate ambiguities, Limit resource consumption, finite set of features and capabilities, follow security and privacy best practices subject to group feedback.

This very generic RESOLUTION is fine and does not directly touch my concern. 

In general, different people have a different understanding what is understood  by "interoperability". Especially, in IoT with its different domains there is not an easy answer to that. I already shared my view on this here:

https://github.com/w3c/wot-profile/issues/25#issuecomment-669878424

As you can see, I'm agree that we have a support for an OotBI Profile such as for Thing2Browser and Thing2Cloud scenarios. 

However, we should make clear in the FPWD, that this is a kind of "offer" for developer that can be followed to possible the interoperability in your  there may other Profiles that  

what is missing that "core" is one profile of man  it should clear in specification that it should be not the core for all kind of WoT scenarios. 
I just to want make clear in the current document that the "core" is not a mandatory thing which have to be NOT implemented when you prefer a different profile, which is, e.g., more relevant for healthcare scenarios and

sebastiankb:
@mlagally 
>We are mixing two topics in this conversation, the profile name and the profile scope and use cases.
Let's treat them separately to make sure we address the individual qustions.

+1

I think, for the name issue we already clarified this yesterday and the corresponding editorial note is included. Regarding the name I already shared my view in July on this:

https://github.com/w3c/wot-profile/issues/21

>RESOLUTION: Accepting the following requirements for the FPWD: Interoperability, Limit and reduce complexity, Eliminate ambiguities, Limit resource consumption, finite set of features and capabilities, follow security and privacy best practices subject to group feedback.

This very generic RESOLUTION is fine, but does not touch my concern directly.

In general, different people have a different understanding what is understood  by "interoperability". Especially, in IoT with its different domains there is not an easy answer to that. I already shared my view on this here:

https://github.com/w3c/wot-profile/issues/25#issuecomment-669878424

As you can see, I'm agree that we have a support for an OotBI Profile such as for Thing2Browser and Thing2Cloud scenarios. I think all thing vendors who want to launch such new devices with this purpose in mind will be glad to know that there is some sort of profile that can be followed to enhance the OotBI. However, we also have other scenarios such as Thing2Thing or Thing2Edge where the “core” profile does not make sense since there are different data model and protocol constraints. I just want to point out that “core” is only one of many profiles and consumer shall NOT be forced to implement always the “core” when it does not make sense in there context / domain.

sebastiankb:
@mlagally
> Is there a description of the four scenarios you refer to and their requirements?

Actually, there are more:

https://www.w3.org/TR/wot-architecture/#sec-architecture-overview 

Expected requirements (e.g. for protocols usage):

https://www.w3.org/TR/wot-architecture/#sec-binding-templates
 
>I'm not sure I agree to your view that a constrained profile does not make sense for Thing2Thing and Thing2Edge scenarios.

I did not write that! 
And again, I support the profile concept. However, I do not support a "core" profile that have to be always part of other (future) profiles.

PR #48, opened 2020-10-09
Some text improvements

sebastiankb:
based on discussion in #47

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/48.html" title="Last updated on Oct 15, 2020, 6:06 AM UTC (ba484af)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/48/bb4847e...ba484af.html" title="Last updated on Oct 15, 2020, 6:06 AM UTC (ba484af)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -97,7 +97,7 @@
 <body>
   <section id="abstract">
     <p>
-        This document introduce a WoT profile concept that supports the so-called <i>out of the box interoperability</i>
+        This document introduces a WoT profile concept that supports the so-called <i>out of the box interoperability</i>
         among WoT-enabled things / devices. A WoT profile defines set of guidelines and rules that can be followed to ensure 
         interoperability. 
     </p>
--- a/index.html
+++ b/index.html
@@ -116,7 +116,7 @@
         profile in an unambiguous way. This mechanism can be
         used to define additional profiles.
       </li>
-      <li>It defines a first base <strong><a>Core
+      <li>It defines a first <strong><a>Core
             Profile</a></strong> of the Thing Description for use with
         the HTTP(S) and selected sub-protocols (e.g., for eventing). 
       </li>
--- a/index.html
+++ b/index.html
@@ -97,13 +97,13 @@
 <body>
   <section id="abstract">
     <p>
-      The present document defines a WoT Profile that enables out of the box interoperability
-      among things and devices. It defines a set of constraints and rules that compliant
-      thing descriptions have to adopt to ensure interoperability.
+        This document intorduce a WoT profile concept that supports the so-called <i>out of the box interoperability</i>
+        among WoT-enabled things / devices. A WoT profile defines set of guidelines and rules that can be followed to ensure 
+        interoperability. 
     </p>
     <p>
-      These rules are prescriptive, to ensure that
-      compliant implementations satisfy the semantic guarantees,
+      These guidelines and rules are prescriptive, to ensure that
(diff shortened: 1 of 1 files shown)


sebastiankb:
@mlagally Please can you share the source of the figures.

index.html
```diff
@@ -97,13 +97,13 @@
 <body>
   <section id="abstract">
     <p>
-      The present document defines a WoT Profile that enables out of the box interoperability
-      among things and devices. It defines a set of constraints and rules that compliant
-      thing descriptions have to adopt to ensure interoperability.
+        This document introduce a WoT profile concept that supports the so-called <i>out of the box interoperability</i>
```
danielpeintner, line 100: "introduce" -> "introduces"

index.html
```diff
@@ -116,25 +116,33 @@
         profile in an unambiguous way. This mechanism can be
         used to define additional profiles.
       </li>
-      <li>In addition it defines a <strong><a>Core
+      <li>It define a first base <strong><a>Core
```
danielpeintner, line 119: "define" --> "defines"

index.html
mlagally, line 100: This language is less clear than the previous text and waters down the purpose of the core profile.

index.html
mlagally, line 119: please clarify what you mean with "a first base"

index.html
```diff
@@ -116,25 +116,33 @@
...
             Profile</a></strong> of the Thing Description for use with
-        the HTTP(S) and selected notification protocols. The <a>Core
-          Profile</a> formalizes the results of several PlugFests
-        that were conducted by the WoT Interest Group and of
-        tests that were conducted as part of the development. It
-        is expected that additional profiles for thing templates
-        and other protocols will be defined in the near future.
+        the HTTP(S) and selected sub-protocols (e.g., for eventing).
```
mlagally, line 121: Not sure we should talk / or select subprotocols. Current proposed event mechanisms include SSE and Longpoll.
These just use HTTP.

mlagally (changes requested):

index.html
sebastiankb: SSE and Longpoll are sub-protocols of HTTP. This is aligned with the current TD spec. Please also see https://www.w3.org/TR/wot-thing-description/#form

index.html
sebastiankb: What you call as "Core" is the first introduced Profile in the specification, right?  So it is the "first base". 
However, I have no hard opinion about that. We can change this back if you want.

index.html
sebastiankb: done

index.html
sebastiankb: done

index.html
sebastiankb: I would say this is the same content. However, I changed the word "introduces" to "defines". What do you think?

mlagally:
@sebastiankb The sources of the drawio files are embedded in the svg files.

sebastiankb:
@mlagally 
>The sources of the drawio files are embedded in the svg files.

I only see this here https://github.com/w3c/wot-profile/tree/master/images

Where are the drawio files?

index.html
sebastiankb: I removed the 'base' term in the text

mlagally:
@sebastiankb agreed to close without merging (Arch call on 15.10)

Issue #55, opened 2020-11-19
Canonical Representation / Signing requirements


mlagally:
We need to review JSON signing mechanism wrt. canonicalisation requirements.

https://tools.ietf.org/html/rfc7515

Check for number representation formats, whitespace handling, sort order.
we should however not duplicate requirements in the profile that are already covered by referenced material.

mlagally:
https://www.w3.org/TR/vc-data-model/ Verifiable Credentials Data Model 1.0 REC

mlagally:
The canonical representation as defined by https://tools.ietf.org/html/rfc8785 looks like a good start.

We shoud check if there are any requirements that require TD specific additional clarifications.

mmccool:
So another thing we should clarify in the TD spec are arrays-of-one-element being expressed as a single value.  I think profiles currently require these to be expressed as arrays but this goes against certain things where we say otherwise, ie security (where the array form, ie with more than one element, is deprecated, since we want to remove the array completely in a future version...).

mmccool:
Note there are already two issues on this:
* https://github.com/w3c/wot-thing-description/issues/940
* https://github.com/w3c/wot-security/issues/166
The one in security points at the TD issue and that's where I think this discussion belongs, as canonicalization requirements (and the scheme for embedding the signing information, eg an ld-proof block) should go into the TD spec, I feel (not profiles; signing should be a universal TD feature, IMO).

mlagally:
Arch call on 10.12.:
We prefer shortness of a TD over verbosity:
- default values MUST not be included in the canonical form of a TD
- prefixes: use the compact form

We have to select a specific signing mechanism / algorithm, may need to update to a new version when an algorithm gets broken.
There needs to be an extension of the TD data model to support https://tools.ietf.org/html/rfc7515.

As a strawman we agree to use JWE with a selected set of algorithms (t.b.d: select algorithms that have not been compromised yet and can be done on resource constrained devices)

See: https://datatracker.ietf.org/wg/jose/documents/

Review the algorithm choices done by CBOR (COSE) 
https://tools.ietf.org/html/rfc8152

mlagally:
We want to reference the corresponding TD chapter and provide required additional constraints, if necessary.

mmccool:
The above PR is readt for review, and should address this issue.

PR #57, opened 2020-12-10
Canonical representation 2

mlagally:
Chapter structure skeleton

targets to address issue https://github.com/w3c/wot-profile/issues/55

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/57.html" title="Last updated on Mar 5, 2021, 1:57 PM UTC (458d6f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/57/f9b6154...458d6f5.html" title="Last updated on Mar 5, 2021, 1:57 PM UTC (458d6f5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -97,18 +97,18 @@
 <body>
   <section id="abstract">
     <p>
-      The <em>WoT Profile Specification</em> defines a Profiling Mechanism and a <em>WoT Core Profile</em>, 
+      The <em>WoT Profile Specification</em> defines a Profiling Mechanism and a <em>WoT Core Profile</em>,
       which enables <em>out of the box interoperability</em>
       among things and devices. <em>Out of the box interoperability</em> implies,
-      that devices can be integrated into various application scenarios without 
-      deep level adaptations. Typically only minor configuration operations are 
-      necessary (such as entering a network key, or IP address) 
-      to use the device in a certain scenario. 
+      that devices can be integrated into various application scenarios without
+      deep level adaptations. Typically only minor configuration operations are
+      necessary (such as entering a network key, or IP address)
+      to use the device in a certain scenario.
       These actions can be done by anyone without specific training.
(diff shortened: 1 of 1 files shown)


mmccool:
Probably should retire this PR, since canonicalization is now in the TD spec, but I will review first to see if any content (eg use case discussion) should be moved over to the TD spec.

mmccool:
Should replace (if necessary) with a link to the Canonical TD definition in the TD spec.  Note however that due to sorting a canonical TD may have to be loaded into memory before processing (we considered but rejected a feature to sort definitions so they appear before uses... it complicated the definition a lot).

mmccool:
Two points discussed in this PR which are not in the TD version:
- Base URLs (not sure why this is an issue)
- Encrypted TDs (this is orthogonal to the issue of canonicalization, but is related to signing; of course a signed TD may not be encrypted, but a public-key-encrypted TD would allow validation that it was encrypted by the owner of the private key).

mmccool:
- Regarding "base":  We could forbid it in canonical form, forcing expression of all URLs in absolute form, but that seems overly draconian and makes chaining signed changes to the base inconvenient/impossible.   However I realized not allowing changes to the base and the endpoint URLs was just a special case of an assertion I already had not allowing changes to how URLs are expressed.  I made this explicit here: https://github.com/w3c/wot-thing-description/pull/1129, please review.
- Regarding encryption: I don't think this belongs under canonicalization, but rather under a discussion of signing, or possible under a discussion of object security for TDs in discovery.  So I left this out of my version.

mlagally:
vF2F on 23.6.:
Canonical forms and signing are targeted for the TD.
This PR needs to be adapted to reflect the state in the TD.
Also some parts wrt. to canonical TDs should be removed.

PR #70, opened 2021-02-25
device categories - initial draft

mlagally:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/70.html" title="Last updated on Mar 5, 2021, 1:57 PM UTC (c88bf91)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/70/594c374...c88bf91.html" title="Last updated on Mar 5, 2021, 1:57 PM UTC (c88bf91)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -595,6 +595,32 @@
     </section>
   </section>
 
+  <section id="device-categoriesdevices">
+    <h2>Device Categories</h2>	  
+<p>
+  In a deployment of WoT conforming to the <a href="https://www.w3.org/TR/wot-architecture/#architecture-abstract">WoT Abstract Architecture</a>
+  we see different device types with typical characteristics.
+  They range (sorted in the order of footprint and capabilities) 
+  from small embedded <em>node</em> devices to <em>gateways</em> or <em>hubs</em> 
+  to powerful <em>edge</em> devices and <em>cloud</em> servers.
+  Interoperability between these devices implies that a core set of features and functionalities
+  is available on <em>all</em> of them.
+</p>
+
+<p>The following device categories describe the footprint and characteristics of typical representants of these classes.
+ This is used to identify the possible features and use cases for these device classes.
+
+ <ul>
+ <li>Node - endpoint with typically 64-128 KB of RAM</li>
+ <li>Gateway/Hub - intermediate node, typically a few MB of RAM</li>
+ <li>Edge - </li>
+ <li>Cloud - multiple CPUs, virtualisation, large storage</li>
+ </ul>
+
+</p>
+  </section>
+
+
 
   <!-- <a>Core Profile</a> -->
 
--- a/index.html
+++ b/index.html
@@ -626,13 +626,13 @@ Note that this is not a strict categorisation, categories may overlap and not al
   <td>Class-0, C0</td>
   <td><< 10 KiB</td>
   <td><< 100 KiB</td>
-  <td>sensor nodes</td>
+  <td>unsecure sensor nodes</td>
 </tr>
 <tr>
   <td>Class-1, C1</td>
   <td>~ 10 KiB</td>
   <td>~ 100 KiB</td>
-  <td></td>
+  <td>secure sensor node</td>
 </tr><tr>
   <td>Class-2, C2</td>
   <td>~ 64 KiB</td>
--- a/index.html
+++ b/index.html
@@ -608,14 +608,62 @@
 </p>
 
 <p>The following device categories describe the footprint and characteristics of typical representants of these classes.
- This is used to identify the possible features and use cases for these device classes.
+ This is used to identify the possible features and use cases for these device classes.</p>
 
- <ul>
- <li>Node - endpoint with typically 64-128 KB of RAM</li>
- <li>Gateway/Hub - intermediate node, typically a few MB of RAM</li>
- <li>Edge - </li>
- <li>Cloud - multiple CPUs, virtualisation, large storage</li>
- </ul>
+ <p>
+We are aligned with the classes defined in https://tools.ietf.org/html/rfc7228 for constrained devices,
+however put the bounds on typical sizes of RAM / Flash in the real world.
+Note that this is not a strict categorisation, categories may overlap and not all memory may be available for user applications.</p>
+
+<table>
+<tr>
+  <th>Category</th> 
+  <th>data size (RAM)</th>
+  <th>code size (Flash, ROM, ...)</th>
+  <th>Typical Representant</th>
+</tr>  
+<tr>
+  <td>Class-0, C0</td>
+  <td><< 10 KiB</td>
+  <td><< 100 KiB</td>
+  <td>sensor nodes</td>
+</tr>
+<tr>
+  <td>Class-1, C1</td>
+  <td>~ 10 KiB</td>
+  <td>~ 100 KiB</td>
+  <td></td>
+</tr><tr>
+  <td>Class-2, C2</td>
+  <td>~ 64 KiB</td>
+  <td>~ 256 KiB</td>
+  <td>nodeMCU</td>
+</tr>
+</tr><tr>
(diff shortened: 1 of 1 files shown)


mlagally:
Arch call on 25.2.
Node terminology discussion:

- constrained node
- sensor node
- directly associated with a sensor or actuator

small, mid size, large category, roles can overlap
constrained 

BACnet controller - S7, couple of MB?

Let's align with building on top of IETF's class definitions RFC-7228:
n= 50 kb class defined by IETF
Class-n+1 = 64KB - 128KB
Class n+2 = 128KB - 512KB
Class n+3 = 512KB -

mlagally:
Classes 0 and 1 are identical to IETF's definition.

Other device examples:
https://customer.honeywell.com/resources/techlit/TechLitDocuments/31-00000s/31-00100.pdf
CPUEach controller uses a 32 bit ATMEL ARM 7 microprocessor.Memory CapacityFlash Memory: 512 kilobytes. The controller is able to retain Flash memory settings for up to ten (10) years.RAM: 128 kilobytes

https://www.w3schools.com/bootstrap/bootstrap_grid_small.asp there is some definition for device classification based on the screen size (but I personally think we should rather use more neutral term like "class")

index.html
```diff
@@ -595,6 +595,32 @@ <h3 id="profile-description-methodology">Methodology</h3>
     </section>
   </section>
 
+  <section id="device-categoriesdevices">
+    <h2>Device Categories</h2>	  
+<p>
+  In a deployment of WoT conforming to the <a href="https://www.w3.org/TR/wot-architecture/#architecture-abstract">WoT Abstract Architecture</a>
+  we see different device types with typical characteristics.
+  They range (sorted in the order of footprint and capabilities) 
+  from small embedded <em>node</em> devices to <em>gateways</em> or <em>hubs</em>
```
sebastiankb, line 604: **small** does not imply resource constrained. **Constrained** or **resource constrained** embedded devices (like field devices) is more common.

index.html
```diff
@@ -595,6 +595,32 @@ <h3 id="profile-description-methodology">Methodology</h3>
...
+  to powerful <em>edge</em> devices and <em>cloud</em> servers.
+  Interoperability between these devices implies that a core set of features and functionalities
+  is available on <em>all</em> of them.
+</p>
+
+<p>The following device categories describe the footprint and characteristics of typical representants of these classes.
+ This is used to identify the possible features and use cases for these device classes.
+
+ <ul>
+ <li>Node - endpoint with typically 64-128 KB of RAM</li>
```
sebastiankb, line 614: **node** only is unspecific. **resource constrained** node 

>typically 64-128 KB of RAM

 I'm not sure about such a range. I would simple say typically KB of RAM

sebastiankb (changes requested):

sebastiankb:
I'm not 100% sure why such a generic section have to be introduced in the core profile. It should be more provided in the architecture document.  On the other hand, WoT claims to be a kind of hardware and platform independence.  Thats the reason why WoT is so interesting! So I'm not sure if we need such text at all. 

If this is the reason for the motivation to limit the TD size, then we must first discuss the concrete scenarios as already mentioned [here](https://github.com/w3c/wot-profile/pull/69#discussion_r578584893).

benfrancis:
@mlagally What are you hoping to achieve by defining these classes of device? It all seems arbitrary to me and of little practical use. Hardware is always evolving and the software/firmware running on a device in addition to any software that might be used for processing Thing Descriptions will vary greatly.

It may be useful to provide a separate profile for constrained devices which uses a different protocol and serialisation format and imposes additional constraints, but given the number of variables in terms of resource needs, is it not better to leave it up to developers which profile is best suited to their particular use case?

mlagally:
@benfrancis 
In recent architecture calls we had quite some conversation cycles because of using terms like ESP, Raspberry, Gateway, Node MCU, ...
In a W3C spec we cannot make reference to products and rather want to talk about capabilities of devices.
This classification helps to ask questions like: Can a TD consumer be implemented in a Class-X devices, and if yes, what kind of constraints does that imply.
One possible outcome of the discussion is to exclude some classes from the consumer role.

benfrancis:
> This classification helps to ask questions like: Can a TD consumer be implemented in a Class-X devices, and if yes, what kind of constraints does that imply.

I just don't think it's possible to generalise answers to these types of questions. An HTTP+JSON profile could be used by a $5 microcontroller while a $5m satellite with a state of the art multi-core processor may want to use a CoAP+CBOR profile due to low network speeds/high latency. It really depends on the individual use case.

> One possible outcome of the discussion is to exclude some classes from the consumer role.

I'm surprised that you use the consumer in this example, as I would have though that it's the producer which is more likely to be a constrained device, which just further demonstrates how difficult it is to provide general answers to such questions.

I would instead propose that it is enough to assume that many WoT devices (and gateways, and clients) could support an HTTP+JSON profile with a large amount of human readable metadata, but some WoT devices (and clients) may be better suited to a CoAP+CBOR profile with more condensed metadata. It would then be up to developers to choose which profile is the best fit for their implementation.

mlagally:
@benfrancis 
The initial questinon that triggered the discussion was: What contraints do we have to consider if we implement a **consumer** on a resource constrained device. A deployment scenario is a connected set/network of sensors in a mesh.
There may be a gateway, but it could be only offering proxy functionality to the outside world, no directory, just P2P discovery - you could think of a UPnP like discovery / usage scenario. 

This is a P2P deployment where a Class-2-3 device consumes the TD of another Class 2-3 device to receive telemtry data or control the device.

One comment about your remark: 
"It would then be up to developers to choose which profile is the best fit for their implementation."

The profile work targets business customers, who select off-the-shelf devices that fit with their needs, i.e. they guarantee to work in their environment for their intended use cases. This is the core point of the profile: Have devices that have a guaranteed deterministic behavior within assured operation ranges. It is a bit like the MIL standard for chips that guarantees that compliant devices operate within specified temperature bounds. 
In our case these limits are size and complexity.

The development model is that individual devices are built by different companies who do not know each other, i.e. use only the TD for communicating about their capabilities.
The end user is NOT a developer who knows the difference between CBOR and JSON but just inspects the label on the device carton to find out if the device works for him.

benfrancis:
@mlagally wrote:
> What contraints do we have to consider if we implement a consumer on a resource constrained device.

OK, thank you for explaining where the requirement comes from. I guess I just haven't come across a use case where the consumer has more constrained hardware than the producer, at least not to the extent that the size of a Thing Description would be the limiting factor. The firmware/software stack needed to fetch and parse _any_ TD and send and receive data from the device it describes seems likely to use an order of magnitude more resources than the metadata describing the device's capabilities (its TD).

> A deployment scenario is a connected set/network of sensors in a mesh.

Is this an actual scenario that has come up in PlugFests, or just a theoretical one? The Web of Things seems ill suited to implementing a mesh network of constrained devices (there are much better protocols available for that like Zigbee/Z-Wave/Bluetooth/Thread), but it can act as a web-based abstraction layer above the mesh networking layer using a client-server model.

I'm not assuming that a gateway is always involved, but I do think that in practice it's likely that most WoT consumers are either a gateway, cloud service or a desktop/mobile application.

Even if we assume that TDs will be used for device to device interactions on constrained devices, I don't see that setting arbitrary limits on the size of a TD really helps much since the size of the TD is only one small factor in the resources needed to communicate with a device.

I understand the problem you're trying to solve, but I really don't believe that the size of Thing Descriptions is what's preventing interoperability. In my opinion we should focus less on adding arbitrary constraints to an abstract data model and more on defining a finite set of protocols and serialisation formats conformant devices must support and how they are used.

mlagally:
Resolution in vF2F on June 23: Retarget to arch specification

mlagally:
This section was retargeted to the architecture spec, this issue can be closed.

Issue #10, opened 2019-09-02
Reduce or remove the constraints in section 7.2 Baseline Information Model


benfrancis:
I suggest that section 4.1.1 Core Data Model should be removed, or merged with the main [WoT Thing Description](https://w3c.github.io/wot-thing-description) specification.

The profile should instead focus on defining constraints for the payloads of HTTP requests as part of the HTTP protocol binding.

benfrancis:
Revisiting this issue following the [First Public Working Draft](https://www.w3.org/TR/2020/WD-wot-profile-20201124).

The current draft says the goals of the WoT Core Profile are:
> - guarantee interoperability among all implementations of the profile.
> - limit the implementation complexity for resource constrained devices.
> - ensure that a thing description is also useful for a human reader.

I do agree that defining constraints in the [WoT Core Data Model](https://www.w3.org/TR/2020/WD-wot-profile-20201124/#wot-core-data-model) section (now section 5.1) could help interoperability, limit implementation complexity and make Thing Descriptions more human readable. For example, limiting the depth of nested data structures and constraining the values of certain members (e.g. limiting units to a single ontology, though the current draft doesn't do this).

However, I find that these requirements often conflict with each other. For example, section "[5.1.2.1 Mandatory Fields](https://www.w3.org/TR/2020/WD-wot-profile-20201124/#mandatory-fields-0)" sets out a list of mandatory members of a Thing Description.

Regarding the first goal, I would argue that none of these fields need to be mandatory in order to guarantee out-of-the box interoperability.
- `id` is not necessary if the URL of the Thing Description can be used as a fallback identifier for a thing
- `title` and `description` are not strictly needed for interoperability
- The `security` member may not be needed for devices which are intentionally publicly available with no authentication
- The rest of the fields are just nice-to-have additional metadata and are also not necessary for interoperability
- (Making these fields mandatory is also made largely pointless by the recommended practice in section 5.1.2.2 of providing empty strings for fields which aren't actually needed.)

Regarding the second goal, making all these members mandatory actually adds to implementation complexity.

Regarding the third goal, I agree that title and description help improve human readability, though a title can be auto-generated from the key of an interaction affordance in the worst case.

I therefore propose that:
1. If the WoT Core Data Model section is not removed then the number of mandatory fields should at least be drastically reduced (either by moving the constraints to the Thing Description specification, or removing them entirely)
[...]

PR #78, opened 2021-04-22
Restructure WoT Core Profile section

benfrancis:
In #73 we have discussed reducing the scope of the Core Profile (name TBD #21) to focus on interoperability and no longer try to meet requirements around human readability and constrained devices (which often conflict with each other).

This PR therefore proposes modifying the WoT Core Profile to:
1. Remove the WoT Core Data Model section for the reasons described in https://github.com/w3c/wot-profile/issues/73#issuecomment-805010541 and https://github.com/w3c/wot-profile/issues/10
2. Remove the External TD representations section since JSON serialisation, optional JSON-LD semantic annotations and canonicalization will be defined in the Thing Desciption specification
3. Add new sub-sections to the WoT Core Profile:
    1. Identifier - an identifier to denote that a given Thing Description conforms to the Core Profile
    2. Links - link relation types which conformant Web Things MAY use and conformant Consumers MUST support
    3. Security - security schemes which conformant Web Things MAY support and conformant Consumers MUST support
    4. Discovery - discovery mechanisms which conformant Web Things MAY support and conformant Consumers MUST support

A follow-up PR would be required to update the Introduction, Conformance and Profiling Mechanism sections to reflect these changes.

Note: This is a companion to #77 which proposes a new Protocol Binding section

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/78.html" title="Last updated on Jul 22, 2021, 10:24 AM UTC (329dcc5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/78/6ca51a8...benfrancis:329dcc5.html" title="Last updated on Jul 22, 2021, 10:24 AM UTC (329dcc5)">Diff</a>


index.html
```diff
@@ -598,870 +529,53 @@ <h3 id="profile-description-methodology">Methodology</h3>
...
-          <h4>Recommended Practice</h4>
-          <p>When using the "no security" or "Basic Auth"
-            security schemes it is strongly recommended to
-            use transport layer encryption.</p>
-        </section>
-      </section>
+    <!-- Identifier -->
+    <section id="identifier">
+      <h3>Identifier</h3>
+      <p class="ednote">
```
sebastiankb, line 565: should be in sync with https://github.com/w3c/wot-thing-description/issues/806

egekorkan:
A question about links, what does it mean to support a link relation type? Any Consumer can go to that link just by following it. Does it mean that if there is a link that means a user interface, the Consumer MUST have a rendering engine to display it?

index.html
benfrancis: If `profile` is defined in the Thing Description specification, do you think this section is still needed in the Profile specification? Or should the Profile specification just refer to the relevant section of the Thing Description specification in its introduction for example?

Is anyone working on a PR for https://github.com/w3c/wot-thing-description/issues/806 yet?

benfrancis:
> A question about links, what does it mean to support a link relation type? Any Consumer can go to that link just by following it. Does it mean that if there is a link that means a user interface, the Consumer MUST have a rendering engine to display it?

That's a good question. Perhaps in that case support should not be mandatory.

In the web interface of WebThings Gateway these types of links are just rendered as an icon with a hyperlink which when clicked opens the custom HTML user interface of a Thing in a new tab, but it's possible some Consumers may not have a user interface at all.

I've updated that text (which is so far just an editor's note) to say MAY instead of MUST.

mmccool:
We probably need to look at links on a case-by-case basis.  Note also that actually following links is a potential privacy risk, so generally we should try to make TDs self-contained.  Note that some links may also point back to the same source as the TD, eg for icons.  These "same-source" links would be safer wrt to privacy.  An alternative would be to embed such resources in the TD somehow, and maybe allow links to use a local json-pointer (of course binary data would need to be base-64 encoded or something).

Hmm.  Maybe directories should cache resources like icons?  I'll have to think about that.

benfrancis:
I have rebased this pull request to resolve merge conflicts and modified it to keep the security section from the Core Data Model section as part of the Core Profile.

@mlagally This pull request is 3 months old today, it would be great to discuss it in today's WoT Architecture call if possible.

mlagally:
@benfrancis 
I already commented on the datamodel section that you deleted in the last architecture call and explained why we need it.

We agreed to prioritize interoperability over human readability, so the normative requirements of that section, that relate to human readability should be moved to the "recommended practice" sections of the corresponding subchapters.
This could be handled in a separate PR.

The additional text you created should be moved into a stand-alone pull request to make sure we can handle it separately and include it into the document in due course.

benfrancis:
@mlagally As discussed on Architecture call on 22nd July, we agreed in #73 to reduce the scope of the Core Profile from its previous goals of:
1. Interoperability
2. Human readability
3. Supporting constrained devices

to focus only on the goal of **interoperability**.

I am arguing that _none_ of the constraints proposed in the current Core Profile Data Model section are needed for interoperability and only contribute to human readability and supporting constrained devices, and therefore the whole section should be removed. If there are any constraints which we agree should apply to _all_ web things, then I am suggesting they should be added to the Thing Description specification rather than the Core Profile.

I previously asked you for specific examples of constraints you felt needed to be kept and I responded to those specific points in https://github.com/w3c/wot-profile/issues/73#issuecomment-805010541, explaining why I don't think they help interoperability and in some cases may even harm interoperability. If you still believe there are specific constraints that should be kept because they contribute to interoperability, can you please either explain that here or by responding in #73.

benfrancis:
> The additional text you created should be moved into a stand-alone pull request to make sure we can handle it separately and include it into the document in due course.

I also explained on the call that it doesn't really make sense for the proposed new sections to be added in a separate PR, because they don't make sense without the other changes. E.g.:
1. The Links section contradicts current text in the Data Model section
2. The Security section replaces current text in the Data Model section

index.html
```diff
@@ -516,996 +448,45 @@ <h2>Profiling Mechanism</h2>
...
-            ,
-            <code>oct</code>
-            or
-            <code>bin</code>
-            in this case to achieve interoperability.
-          </p>
-        </section>
-      </section>
-
+    This section defines the <a>Core Profile</a> by describing:
```
mlagally, line 454: The new language should be added, however removing the core data model is not appropriate, since it will break profile use cases.

index.html
```diff
@@ -416,60 +402,6 @@ <h4 id="additional-definitions-">Additional Definitions:</h4>
       <dd>
         The Web of Things Thing Description as defined in [[wot-thing-description]].
       </dd>
```
mlagally, line 404: Agree to remove.

index.html
```diff
@@ -386,18 +384,6 @@ <h4 id="additional-definitions-">Additional Definitions:</h4>
       <dd>
         Synonym for <a>WoT Core Profile</a>.
       </dd>
```
mlagally, line 386: agree to remove

index.html
```diff
@@ -198,8 +198,6 @@ <h3>Motivation for a Profile</h3>
   <section id='sotd'></section>
 
   <section id="intro">
```
mlagally, line 200: Why do you want to remove a h1 heading?

index.html
mlagally: We discussed the use cases and motivation for the common data model in the October vF2F. Please see presentation slides: https://github.com/w3c/wot/blob/main/PRESENTATIONS/2021-10-online-f2f/2021-10-27-WoT-F2F-Architecture%2BProfiles-Lagally.pdf

index.html
```diff
@@ -516,996 +448,45 @@ <h2>Profiling Mechanism</h2>
...
-          <p>When using the "no security" or "Basic Auth"
-            security schemes it is strongly recommended to
-            use transport layer encryption.</p>
-        </section>
-      </section>
+    <!-- Identifier -->
+    <section id="identifier">
+      <h3>Identifier</h3>
+      <p class="ednote">
+        This section will define an identifier which can be included in a
```
mlagally, line 486: text should be added

index.html
```diff
@@ -1927,79 +908,48 @@ <h4>Error Responses</h4>
...
+      <h3>Links</h3>
+      <p class="ednote">
+        This section will define link relation types which conformant Web Things
+        MAY use and conformant Consumers MAY support
+        (e.g. <code>{ "rel": "alternate", "type": "text/html"}</code> to link
+        to an HTML user interface).
       </p>
+    </section>
 
-        <section id="canonical-representation">
```
mlagally, line 1943: We need the section on canonical representation, since this may be removed from the TD.

index.html
```diff
@@ -1927,79 +908,48 @@ <h4>Error Responses</h4>
...
-          </p>
-	  <p>
-	     The canonical JSON representation format of a TD adopts the JSON Canonicalization Scheme (JCS) defined by
-	     <a href="https://tools.ietf.org/html/rfc8785>RFC-8785</a> including errata <a href="https://www.rfc-editor.org/errata/eid6292">6292</a>.
-	  </p>
-      </section>
-
-
+    <!-- Security -->
+    <section id="security">
```
mlagally, line 923: We need a security section, however need to discuss the supported mechanisms under implementation constraints for plugfest members.

I suggest to exclude "no sec" and require basic auth as the minimum. 
The other mechanisms need input from plugfest members.

index.html
```diff
@@ -1927,79 +908,48 @@ <h4>Error Responses</h4>
...
-		</section -->
-  </section>
-  <section id="open-issues">
-    <h3>Open Issues</h3>
-    <ul>
-      <li>define action semantics</li>
-      <li>define subscription mechanism</li>
-      <li>Examples</li>
-    </ul>
+    <!-- Discovery -->
```
mlagally, line 945: I don't think we will be able to define a discovery mechanism in the first profile release.

mlagally (changes requested):

mlagally:
I reviewed this MR again, which has a lot of valuable parts.
However merging it as a whole does not work, since it contains too many contentious sections.

To move forward I recommend to split it as described in the individual comments.

benfrancis:
OK, I will split this PR out into smaller ones and close this one once that's done.

index.html
benfrancis: @mlagally I apologise for not being able to attend that meeting due to illness, but I see nothing in that presentation which convinces me that the data model constraints proposed in the current WoT Core Data Model section are necessary, or helpful. Please see my extensive comments in https://github.com/w3c/wot-profile/pull/93#pullrequestreview-717982575 from July in which I reiterate why I think they are undesirable and in some cases conflict with the agreed goal of the specification. Ege has said he agrees with those comments, and I note that you haven't yet responded to them.

I get the impression that you are the only member of the group who thinks that including these data model constraints in the Core Profile is a good idea, and that the majority of members of the group actually think that the majority of the proposed constraints aren't necessary. I am happy to continue to discuss this further to try to reach a consensus, but if we can't reach a consensus then according to the [Working Group's decision policy](https://www.w3.org/2020/01/wot-wg-charter.html#decisions), I understand the Chairs may call for a group vote.

index.html
benfrancis: It's already commented out, this is just cleanup :)

index.html
benfrancis: What is the canonical representation used for in the WoT Profile specification, and why do you think it may be removed from the Thing Description specification?

index.html
benfrancis: I agree this requires some discussion, but I strongly object to excluding the `nosec` security scheme. This is currently used by all [16 web thing libraries](https://webthings.io/framework/) in the WebThings Framework and there can be good reasons for a web thing not providing its own authentication.

index.html
benfrancis: I've been thinking about this and I agree that might be the outcome.

Of the introduction mechanisms defined by the WoT Discovery specification, I think [DNS-based Service Discovery](https://w3c.github.io/wot-discovery/#introduction-dns-sd) could be a candidate for a mandatory discovery mechanism in the Core Profile, but that may be asking too much.

My current thinking is that  [Direct](https://w3c.github.io/wot-discovery/#introduction-direct) discovery could be made the mandatory discovery mechanism by simply requiring that a Thing Description conforming to the Core Profile must be resolvable through an HTTP URL which can be manually typed or copy & pasted in order to locate the Thing Description.

index.html
benfrancis: (Same with the previous two removals)

benfrancis: https://github.com/w3c/wot-profile/pull/116

index.html
benfrancis: https://github.com/w3c/wot-profile/pull/116

index.html
benfrancis: https://github.com/w3c/wot-profile/pull/116

index.html
benfrancis: https://github.com/w3c/wot-profile/pull/117

benfrancis:
As requested, I have split this PR out into separate smaller PRs:

1. #116
2. #117
3. #121
4. #123
5. #124
6. #125

Whether or not we need a Links section is being discussed in #114.

PR #79, opened 2021-04-22
New Profile Mechanism section

benfrancis:
This PR proposes a new simplified Profile Mechanism section which simply defines a mechanism to denote that a given Web Thing conforms to a particular profile.

As discussed in https://github.com/w3c/wot-profile/issues/73#issuecomment-804252286 and #2

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/79.html" title="Last updated on Jun 23, 2021, 1:25 PM UTC (d1933e4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/79/1580311...benfrancis:d1933e4.html" title="Last updated on Jun 23, 2021, 1:25 PM UTC (d1933e4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -476,6 +476,11 @@
   <section id="profiling-mechanism">
     <h2>Profiling Mechanism</h2>
     <p>
+      In order to conform with a profile, a
+      <a href="https://www.w3.org/TR/wot-architecture/#dfn-thing">Web Thing</a>
+      MUST conform with all the normative statements in the profile's specification.
+    </p>
+    <p>
       In order to denote that a given
       <a href="https://www.w3.org/TR/wot-architecture/#dfn-thing">Web Thing</a>
       conforms to one or more profiles, its Thing Description MUST include a
--- a/index.html
+++ b/index.html
@@ -478,12 +478,12 @@
     <p>
       In order to denote that a given
       <a href="https://www.w3.org/TR/wot-architecture/#dfn-thing">Web Thing</a>
-      conforms to a particular profile, its Thing Description
-      [[wot-thing-description]] MUST include a <code>profile</code> member whose
-      value is set to a valid profile identifier corresponding to that profile.
-    </p>
-    <p>
-      A profile identifier MUST be a valid URI [[RFC3986]].
+      conforms to one or more profiles, its Thing Description MUST include a
+      <a href="https://w3c.github.io/wot-thing-description/#thing">
+      <code>profile</code></a> member [[wot-thing-description]]. The value of
+      the <code>profile</code> member MUST be set to either a valid URI
+      [[RFC3986]] identifying a single profile, or an array of valid URIs
+      identifying multiple profiles.
     </p>
     <pre class="example">
     {
@@ -495,22 +495,13 @@
       ...
     }
     </pre>
-    <p>
-      In order to denote that a given
-      <a href="https://www.w3.org/TR/wot-architecture/#dfn-thing">Web Thing</a>
-      conforms to multiple profiles, its Thing Description
-      [[wot-thing-description]] MUST include a <code>profile</code> member whose
-      value is set to an array of valid profile identifiers corresponding to the
-      profiles to which it conforms.
-      </thead>
-    </p>
     <pre class="example">
     {
       "@context": "https://www.w3.org/2019/wot/td/v1",
       "id": "urn:dev:ops:32473-WoTLamp-1234",
       "profile": [
         "https://www.w3.org/2021/wot/profile/core",
-        "https://www.w3.org/2021/wot/profile/constained"
+        "https://www.w3.org/2021/wot/profile/constrained"
       ],
       "title": "My Lamp",
       "description": "A web connected lamp",
--- a/index.html
+++ b/index.html
@@ -98,7 +98,7 @@
       background-color: rgb(230,230,230)
     }
     </style>
-    
+
 </head>
 
 <body>
@@ -179,7 +179,7 @@
       WoT Thing Description [[wot-thing-description]] define a
       powerful description mechanism and a format to describe
       myriads of very different devices, which may be connected
-      over various protocols. 
+      over various protocols.
       </span>
       The format is very flexible and open
       and puts very few normative requirements on devices that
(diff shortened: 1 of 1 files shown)


benfrancis:
Note: It may make sense to move this section of the WoT Profile specification to the WoT Thing Description specification.

index.html
```diff
@@ -459,146 +459,53 @@ <h4 id="additional-definitions-">Additional Definitions:</h4>
...
-      dependent on the context where the keyword is used. In
-      addition the <a>WoT Thing Description</a> specification defines relationships and
-      corresponding cardinalities between these classes.
+      A profile identifier MUST be a valid URI [[RFC3986]].
     </p>
+    <pre class="example">
+    {
+      "@context": "https://www.w3.org/2019/wot/td/v1",
+      "id": "urn:dev:ops:32473-WoTLamp-1234",
+      "profile": "https://www.w3.org/2021/wot/profile/core",
```
sebastiankb, line 478: profile term does not exist yet, but I would be ok with `profile`. 
https://github.com/w3c/wot-thing-description/issues/806

sebastiankb: still we need a good name for "core"

index.html
```diff
@@ -459,146 +459,53 @@ <h4 id="additional-definitions-">Additional Definitions:</h4>
...
-				HTTP(S). It is envisioned, that future versions of this document
-				contain additional mapping to Web Sockets and other protocols.
-                -->
-      </p>
-    </section>
+    <pre class="example">
+    {
+      "@context": "https://www.w3.org/2019/wot/td/v1",
+      "id": "urn:dev:ops:32473-WoTLamp-1234",
+      "profile": [
```
sebastiankb, line 497: interesting. +1 to define [profile](url) as string and array of strings

index.html
benfrancis: I suggest "Common Profile" https://github.com/w3c/wot-profile/issues/21#issuecomment-806543255

index.html
```diff
@@ -459,146 +459,53 @@ <h4 id="additional-definitions-">Additional Definitions:</h4>
...
+        "https://www.w3.org/2021/wot/profile/core",
+        "https://www.w3.org/2021/wot/profile/constained"
```
egekorkan, line 499: should be constrained :)

index.html
egekorkan: Just to be precise, the text above is read as if it should always be an array

relu91 (approved):
Like the idea of having a dedicated field to declare profile conformance.

danielpeintner:
I know we have some wildlife but we might consider whether to use singular or plural form (`profile` vs `profiles`)

index.html
benfrancis: @sebastiankb: If `profile` is defined in the Thing Description specification, do you think this section is still needed in the Profile specification? Or should the Profile specification just refer to the relevant section of the Thing Description specification in its introduction for example?

Is anyone working on a PR for https://github.com/w3c/wot-thing-description/issues/806 yet?

sebastiankb: yes, I will do it until the next TD meeting. That means we will have it available for the next WD release.

index.html
```diff
@@ -459,146 +459,53 @@ <h4 id="additional-definitions-">Additional Definitions:</h4>
...
-      The <a>WoT Thing Description</a> specification already has some constraints,
-      but there is a wide variety of variations that are left to
-      the interpretation or the discretion of an implementer. The
-      rationale for the <a>Core Profile</a> is not to forbid
-      complex things, rather to enable statements like:
+      In order to denote that a given
+      <a href="https://www.w3.org/TR/wot-architecture/#dfn-thing">Web Thing</a>
+      conforms to multiple profiles, its Thing Description
+      [[wot-thing-description]] MUST include a <code>profile</code> member whose
+      value is set to an array of valid profile identifiers corresponding to the
```
sebastiankb, line 489: I would suggest a `string` only for a single profile. An `array` if multiple profiles are implemented. We have multiple places in the TD for such kind of approach (e.g., `security`, `@type`, ...)

index.html
```diff
@@ -459,146 +459,53 @@ <h4 id="additional-definitions-">Additional Definitions:</h4>
   </section>
 
   <!-- Profiling mechanism -->
-
   <section id="profiling-mechanism">
-    <h2 id="the-profiling-profile">Profiling Mechanism</h2>
+    <h2>Profiling Mechanism</h2>
```
mlagally, line 463: This MR does not contain a new profile mechanism section and it is not obvious 
why the entire profiling mechanism section has been deleted. 

Defining a profiling mechanism is part of the requirements for the profile spec,
so deleting this section is not an option.

index.html
mlagally: @sebastiankb 
Please add a link to the PR that introduces the "profile" identifier in the TD spec to give context.

index.html
mlagally: @sebastiankb The "array" AND "string" type definition is causing confusion now and then - as in the rest of the profile spec, I suggest to describe single value profile values always as arrays.

sebastiankb:
From today's TD call: 
- we have to clarify how the consumer selects the profile when multiple are provided by the producer. 
- can a consumer mix profiles?

index.html
benfrancis: This was what I intended to specify, so I need to revisit the wording to make this more clear.

index.html
benfrancis: `"profile": "https://www.w3.org/2021/wot/profile/core"`

That's the mechanism.

In my opinion this is all that is needed, and even this should probably be moved to the Thing Description specification and just referenced from here.

index.html
benfrancis: @egekorkan Yes sorry that's unclear, I'll work on improving the wording.

@mlagally Given the vast majority of Things are likely to only conform to a single profile, I think that would be wasteful. It's only really complicated if it's not defined from the beginning, which then causes backwards compatibility problems in the future.

index.html
benfrancis: I've added a link to the definition of the profile member in the TD specification now that has landed.

index.html
benfrancis: @egekorkan OK, I've changed the wording, let me know if that's clearer?

index.html
benfrancis: Good catch, fixed.

index.html
benfrancis: Done.

benfrancis:
> * we have to clarify how the consumer selects the profile when multiple are provided by the producer.
> * can a consumer mix profiles?

It's a bit difficult to define that until we know what other profiles may exist, but in general I'd say that Web Things can support multiple profiles, but Consumers should probably just pick one to use based on the protocols they support. There may be exceptions where they want to use multiple profiles, e.g. if they want to use WebSockets for events and HTTP for everything else.

mlagally:
@benfrancis 
I think the new section with the examples adds value and should be factored out into a separate PR, so we can discuss it stand alone.
We the can debate the controversy, whether we should to eliminate the profiling mechanism section, that is a mandate of the charter, separately.

benfrancis:
Hi @mlagally, the charter requires the "Definition of a mechanism to identify the profile or profiles that a given TD is conformant to."

In my opinion this simplified version of the profile mechanism section fulfils that requirement while removing a lot of unnecessary text. The mechanism to identify that a given TD is conformant with a profile is the `profile` member of the Thing Description, which accepts a URI identifying a profile.

Being conformant with a profile simply means conforming with all of the normative statements in its specification. I have added some additional text to make that more clear.

Simply adding more text to what's already there would not fulfil the goal of this PR, which is to simplify the profile mechanism.

Issue #6, opened 2019-08-30
Recommended Security


mmccool:
In section <strike>4.1.9.1</strike> 5.4, Recommended Practice for Security, it currently states

"When using the "no security" or "Basic Auth" security schemes it is strongly recommended to use transport layer encryption."

This should instead state, normatively:

"When using the "Basic Auth" or "Digest Auth" security schemes it is REQUIRED to use transport layer encryption.  For all other schemes use of transport layer encryption is RECOMMENDED."

This is because Basic and Digest Auth are basically useless without transport encryption, they are designed to be used with it: they basically transmit passwords in the clear otherwise.  Worse, using these without encryption gives people a false sense of confidence.

I will have to look at the other schemes, but in general we should say encryption is at least RECOMMENDED for all of them.  

For "nosec" while you are not controlling access you might still want to control confidentiality with transport security, but on balance I think for this we can leave it up to the user; this mode should however be primarily used only for development.   I think it's probably fine to lump it under RECOMMENDED.

Note however that using transport encryption with HTTP in systems not connected to the internet can be a bit troublesome, but oh, well.

benfrancis:
Adding a note here based on [feedback](https://github.com/w3c/wot-profile/pull/78#discussion_r742749861) on https://github.com/w3c/wot-profile/pull/78 that the final list of mandatory security schemes for Core Profile compliant Consumers should be reviewed based on implementation experience.

My view is that this list should be kept relatively short. I advocate for `NoSecurityScheme` and `OAuth2SecurityScheme` being on the list.

mmccool:
Note: I am fine with removing Digest.  It's not any better than basic when TLS is used, basic is easier to implement, and without TLS digest is not secure, either.

mmccool:
Discussion in Security TF 2023.02.13:
- It seems the doc was reorganized so we are assuming this issue is referring to section 5.4.
- It seems Digest has been removed, and the Security TF is ok with that change.
- List of assertions that MAY be supported by Things is redundant, since TDs already allow these.  If, however, the intent is that these are the ONLY security schemes a Web Thing conformant to a profile CAN use, then the assertion should be worded as "A Web Thing MUST use one of the following...".  This makes sense for interoperability.
- Secure transport is not discussed.  It is covered in Architecture, so the only question is whether it would be redundant to mention it here.  Perhaps at least an informative note if it is covered in Architecture.
[...]

PR #87, opened 2021-07-15
Placeholder section on transport security

mlagally:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/87.html" title="Last updated on Jul 20, 2021, 7:35 AM UTC (e677c66)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/87/f1ec8e2...e677c66.html" title="Last updated on Jul 20, 2021, 7:35 AM UTC (e677c66)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/87.html" title="Last updated on May 23, 2022, 10:22 AM UTC (e677c66)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/87/f1ec8e2...e677c66.html" title="Last updated on May 23, 2022, 10:22 AM UTC (e677c66)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1925,6 +1925,18 @@
           </span>
         </p>
       </section>
+
+      <section id="security">
+        <h4 id="security">Transport security</h4>
+      </p>
+      <p class="ednote">
+        Consider mandating HTTPs, at least for things that are used outside of a closed network.
+        Some other standards, such as Echonet use HTTPS only.
+      </p>
+  
+      
+      </section>
+
     </section>
 
     <!-- External Representation -->


index.html
```diff
@@ -1925,6 +1925,18 @@ <h4>Error Responses</h4>
           </span>
         </p>
       </section>
+
+      <section id="security">
+        <h4 id="security">Transport security</h4>
+      </p>
+      <p class="ednote">
+        Consider mandating HTTPs, at least for things that are used outside of a closed network.
+        Some other standards, such as Echonet use HTTPS only.
```
danielpeintner, line 1934: nit:
* once "HTTPs" while the second time "HTTPS" with capital S letter
* ECHONET seems to be always capitalized

benfrancis:
What would mandating HTTPS mean in practice and what's the definition of a closed network? Is the idea that a conforming WoT consumer would simply have to refuse to communicate with a WoT device which doesn't use TLS? This seems a bit heavy-handed to me given there may be valid use cases where plain HTTP may be unavoidable (e.g. for local domains or constrained devices lacking the processing power for cryptography algorithms).

Perhaps a more balanced approach would be to strongly _recommend_ the use of TLS and encourage user agents to inform users of insecure connections like web browsers do.

benfrancis (commented):
What would mandating HTTPS mean in practice and what's the definition of a closed network? Is the idea that a conforming WoT consumer would simply have to refuse to communicate with a WoT device which doesn't use TLS? This seems a bit heavy-handed to me given there may be valid use cases where plain HTTP may be unavoidable (e.g. for local domains or constrained devices lacking the processing power for cryptography algorithms).

Perhaps a more balanced approach would be to strongly recommend the use of TLS and encourage user agents to inform users of insecure connections like web browsers do.

benfrancis:
See also:
- https://github.com/w3c/wot-discovery/issues/139
- https://github.com/w3c/wot-security/issues/185#issuecomment-836138091
- https://github.com/w3c/wot-profile/issues/6

benfrancis:
Note that one of the changes proposed in #78 is to add a security section to the Core Profile which defines security schemes which conformant Web Things MAY support and conformant Consumers MUST support. That goes beyond just transport security.

index.html
mlagally: 👍

mlagally:
> Note that one of the changes proposed in #78 is to add a security section to the Core Profile which defines security schemes which conformant Web Things MAY support and conformant Consumers MUST support. That goes beyond just transport security.

The text in #78 is basically just a heading and a one liner. You could create a dedicated PR for that.
However we need a more specific security section that defines authentication, transport security etc.
I was counting on @mmccool  and the Security TF to help us here.

mmccool:
See also https://github.com/w3c/wot-security-best-practices/issues/13 which captures some discussions we have been having in the Security TF about local transport security (e.g. https on the local network) which is the hard problem here.  Summary: it's hard to mandate use of https in use cases like local networks in the home, unfortunately.

mmccool (commented):
I don't think mandating HTTPS is *feasible* in a local environment, unfortunately.  The situation is quite complicated, and I would recommend instead that we mandate "best practices" for security and defer to the WoT Security Best Practices document.

Generally: Security is important but does not directly impact interoperability, which should be our focus in profiles.

mmccool:
Don't think we can mandate HTTPS, so... I think we should simply defer to the security best practices document (which we are working on actively right now, and want to publish in the fall.)  Note however it is informative, so the reference in "profiles" should be an informative recommendation.

Do we have anything around security that affects interop?   We do need to specify a specific set of security schemes.   We probably do want to review the list and perhaps constrain them a bit.   APIkey in particular is quite open-ended.   But even OAuth2 has some variations e.g. on where the token is embedded.   We should also perhaps disallow certain things like "in": "url" which is generally a bad idea, but is there for compatibility reasons (certain brownfield devices).

OliverPfaff:
Re mandating "HTTP-over-TLS - yes/no/maybe?": it should be considered that NETCONF (resp. RESTCONF) probably is a closer "neighbor" of WoT than HTTP with its origins and majority usage in IT ecosystems

NETCONF mandates security in RFC 6241 (see https://datatracker.ietf.org/doc/html/rfc6241#section-2.2). NETCONF applies a security-always-on paradigm while Web applies a security-as-an-option paradigm

From my perspective WoT security should make an explicit decision about the designated security paradigm in e.g. {security-as-an-option (opt-in), security-by-default (opt-out), security-always-on (no opt-out)}

benfrancis:
See #124

PR #91, opened 2021-07-21
Example TD and declarative protocol binding illustration - closes #90

### LINKED ISSUE #90: Provide example TD and declarative protocol binding illustration for Core Profile
Provide an example Thing Description at the start of the Core Profile Protocol Binding chapter, and then an informative canonical Thing Description with all of the protocol details from the protocol binding applied at the end.

This will help demonstrate how a Web Thing conforming to the Core Profile can provide a simple Thing Description, which a conforming Consumer can then expand out into a more comprehensive Canonical Thing Description by applying all the details from the protocol binding specification (e.g. ops, HTTP headers, data schemas, response codes etc.).

benfrancis:
This is an example Thing Description and corresponding declarative protocol binding to help illustrate the Core Profile Protocol Binding, as suggested in #90.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/91.html" title="Last updated on Nov 22, 2021, 6:10 PM UTC (c666c29)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/91/d51d816...benfrancis:c666c29.html" title="Last updated on Nov 22, 2021, 6:10 PM UTC (c666c29)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/91.html" title="Last updated on Mar 28, 2022, 3:15 PM UTC (8f4b9d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/91/0eaf181...benfrancis:8f4b9d1.html" title="Last updated on Mar 28, 2022, 3:15 PM UTC (8f4b9d1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1695,7 +1695,7 @@
         {
           "@context": "https://www.w3.org/2019/wot/td/v1",
           "id": "https://mywebthingserver.com/things/lamp",
-          "profile": "https://www.w3.org/2022/wot/profile/core",
+          "profile": "https://www.w3.org/2022/wot/profile/core/v1",
           "base": "https://mywebthingserver.com/things/lamp/",
           "title": "My Lamp",
           "description": "A web connected lamp",
--- a/index.html
+++ b/index.html
@@ -1722,7 +1722,7 @@
               "unit": "percent",
               "minimum" : 0,
               "maximum" : 100,
-              "forms": [{"href": "./properties/brightness"}]
+              "forms": [{"href": "./properties/level"}]
             }
           },
           "actions": {
--- a/index.html
+++ b/index.html
@@ -3328,33 +3328,36 @@
           ]
         },
         "actionStatusList": {
-          "type": "array",
-          "items": {
-            "type": "object",
-            "properties": {
-              "status": {
-                "type": "string",
-                "enum": [
-                  "pending",
-                  "running",
-                  "completed",
-                  "failed"
-                ]
+          "type": "object",
+          "additionalProperties": {
+            "type": "array",
+            "items": {
+              "type": "object",
+              "properties": {
+                "status": {
+                  "type": "string",
+                  "enum": [
+                    "pending",
+                    "running",
+                    "completed",
+                    "failed"
+                  ]
+                },
+                "error": {
+                  "type": "object"
+                },
+                "href": {
+                  "type": "string",
+                  "format": "uri",
+                  "const": "./actions/fade/{id}"
+                }
               },
-              "error": {
-                "type": "object"
-              },
-              "href": {
-                "type": "string",
-                "format": "uri",
-                "const": "./actions/fade/{id}"
-              }
-            },
-            "required": [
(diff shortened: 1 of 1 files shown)


mlagally:
@benfrancis 
It is very useful to have a complete TD example, I think we have to discuss if we really need two examples.
The canonical form could be the single format we chose as the normative representation format for profile compliant TDs.

benfrancis:
@mlagally Please see #90 for an explanation for the purpose of the two different examples and why both are needed. You indicated in https://github.com/w3c/wot-profile/issues/81#issuecomment-883160715 that you agreed with this general approach.

benfrancis:
I think this is pretty much complete now and ready for review.

To recap, there are two example Thing Descriptions in this PR.

The first Thing Description is provided at the start of the Core Profile Protocol Binding section and provides an example TD for the Web Thing which is used in all the examples in the protocol binding specification text. This Thing Description is in the concise form that a Web Thing conforming to the Core Profile is able to use due to all the defaults and assumptions implied by setting the `profile` member of the TD to the Core Profile URI.

The second example is an illustration of what that Thing Description would look like if the full protocol binding was instead provided declaratively in the Thing Description, without all of the defaults and assumptions implied by the profile member.

The idea is to provide a concise Thing Description at the start to demonstrate what a conforming Web Thing would actually use in order to benefit from all of the defaults and assumptions, then an expanded form at the end which demonstrates what it would look like with all of those defaults and assumptions made explicit rather than being implied by the profile member.

--- 

Some open points for discussion around the second (expanded) example:
1. There's not really a way to provide a declarative schema for the payload of the `readallproperties` and `writeallproperties` operations. In the Thing Description specification it says that these would be a map of all the properties, keyed by property name, but there's not really a way to show that explicitly as far as I can tell.
2. The expanded example provides a `schema` member in `ExpectedResponse` objects, which is not currently included in the Thing Description specification but I'm [expecting it to be added](https://github.com/w3c/wot-thing-description/issues/1053#issuecomment-959452562). There are many parts of the protocol binding which can't be described declaratively without this.
3. The expanded example does not include all the possible error conditions. There are many possible error codes and it gets quite unweildy to list every possible error in every form. Is there a more concise way of expressing this?

benfrancis:
I've updated the PR to move the second example (the declarative protocol binding example) to an appendix, since it is only an informative example to illustrate the benefits of the profiling mechanism and is not required for implementation. I've also improved the descriptions of the two examples because I wasn't happy with the wording.

mlagally:
@benfrancis Please resolve the merge conflict.

benfrancis:
I've resolved the merge conflict, but these examples will require a re-think if #188 lands because events will be split out into a separate profile.

benfrancis:
Replaced by https://github.com/w3c/wot-profile/pull/195

PR #94, opened 2021-07-29
time format - initial draft

mlagally:
***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 13, 2021, 2:04 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fwot-profile%2F58e5d8c4ffaaa7ed60a3f3359623243c7c66f2a8%2Findex.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
Specification: https://rawcdn.githack.com/w3c/wot-profile/58e5d8c4ffaaa7ed60a3f3359623243c7c66f2a8/index.html?isPreview=true&publishDate=2021-09-13
ReSpec version: 26.13.4
File a bug: https://github.com/w3c/respec/
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 2793ms.
    at ExecutionContext._evaluateInternal (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:221:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer/lib/cjs/puppeteer/common/ExecutionContext.js:110:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:276:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:92:18)
    at async Object.generate [as respec] (/u/spec-generator/generators/respec.js:14:44)
    at async /u/spec-generator/server.js:203:44

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/wot-profile%2394.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -570,6 +570,19 @@
         </section>
 
         <section>
+          <h4>Date format</h4>
+          <p class="rfc2119-assertion" 
+          id="profile-date-format-1">
+            All date and time values MUST use the format defined in [ISO 8601].
+          </p>
+          <p class="rfc2119-assertion" 
+          id="profile-date-format-2">
+            Whenever reasonable, all times SHOULD use UTC as the standard time zone.
+        </p>
+        </section>
+
+
+        <section>
           <h4>Length and Value Limits</h4>
           <p>
             The length of
--- a/index.html
+++ b/index.html
@@ -573,12 +573,17 @@
           <h4>Date format</h4>
           <p class="rfc2119-assertion" 
           id="profile-date-format-1">
-            All date and time values MUST use the format defined in [ISO 8601].
+            All date and time values MUST use the canonical dateTime representation format defined in section 3.2.7 of [xmlschema-2].
           </p>
+	  As described by this section the following constraints must be observed:
           <p class="rfc2119-assertion" 
           id="profile-date-format-2">
-            Whenever reasonable, all times SHOULD use UTC as the standard time zone.
+            All dateTime values MUST use UTC as the time zone and use the 'Z' identifier.
         </p>
+          <p class="rfc2119-assertion" 
+          id="profile-date-format-3">
+	    A time value of 24:00 is NOT PERMITTED, the value 00:00 MUST be used instead.
+	</p>		
         </section>
 
 


mmccool:
Note that canonicalization has some specific constraints, as does the TD spec.   I've been basing the constraints in the TD canonical form on the XML canonical form.  These constraints include insisting on zulu time and annotation (Z) - this is basically UTC; not allowing the use of 24 in the hour field, and constraints on the expression of fractional seconds.   It is not, however, critical that the profile follow these constraints, and some are a bit annoying, but they would make them more consistent with canonical form.

Another general thing we could say is that "date and time SHOULD use the same constraints as canonical form (cite section in TD spec)" which would guarantee they are consistent, rather than repeating the constraints in the profile.

mmccool:
Some thoughts:
- Constraining datetime reduces complexity but does not necessarily improve interop (a consumer can just follow the spec, which is clear, if complex)
- If we do accept this, a dependence on canonical form is ok since that part of canonicalization is not likely to go away

mlagally:
@mmccool 
Thanks for these considerations.
I'm seeing a lot of confusion in an international team based on local thinking and use of local times.
We also experienced some meeting schduling problems, so an agreed time representation helps to minimize mistakes here.

I checked the constraints in the canonical section of the TD spec on the dateTime type.
A reader currently needs to go through the XMLSchema spec and read some heavyweight text, and follow on to ISO 8601 before he knows how to represent date and time values. This appears to be a bit counterintuitive - if we are able to give a clear guidance here it would avoid confusion and misinterpretation.

index.html
```diff
@@ -569,6 +569,19 @@ <h4>Mandatory fields</h4>
...
 
+        <section>
+          <h4>Date format</h4>
+          <p class="rfc2119-assertion" 
+          id="profile-date-format-1">
+            All date and time values MUST use the format defined in [ISO 8601].
+          </p>
+          <p class="rfc2119-assertion" 
+          id="profile-date-format-2">
+            Whenever reasonable, all times SHOULD use UTC as the standard time zone.
```
mlagally, line 580: Arch call on 9.9.: We mandate UTC as a timezone using "Z" specifier.

mlagally: normative reference needs to added.

mlagally: Corner cases:
24 hour time corner case is ambiguous, i.e. 24:00, which is 00:00.
We will exclude 24:00
See corresponding section on the canonical form in the TD spec.

PR #119, opened 2021-11-08
Add timestamps to event protocol binding - closes #100

### LINKED ISSUE #100: Events need a way to express timestamps
The event model currently does not define a common way of signaling events with time stamps.

These could be part of the object payload or be a part of the format we do in the id.
A standardized time stamp format on events provides enhanced interoperability.
One apporach could be to define a wrapper object around the event payload.

benfrancis:
This PR adds a recommendation that the `id` field of an event should be set to a timestamp in ISO 8601 format, as a way to include timestamps in events as requested by #100.

I was originally going to suggest a Number representing the number of milliseconds since the epoch [as defined in ECMAScript](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-time-values-and-time-range) (amongst may other programming languages), which is a common practice in Server-Sent Events. However, I'm concerned that may cause confusion given the current text in [5.1.1.2 Date format](https://w3c.github.io/wot-profile/#date-format) requires ISO 8601 format for "all date and time values".

I was initially concerned that using an ISO 8601 timestamp might make it harder for implementations to determine whether a given event occurred after the event identified by the `Last-Event-ID` header when dealing with a dropped connection, but upon closer reading of the specification I think it just tries to find an exact string match rather than calculating greater than or less than, so it shouldn't make any difference.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/119.html" title="Last updated on Nov 16, 2021, 4:57 PM UTC (90dd439)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/119/316ba88...benfrancis:90dd439.html" title="Last updated on Nov 16, 2021, 4:57 PM UTC (90dd439)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/119.html" title="Last updated on Jul 27, 2022, 4:16 PM UTC (be02dc0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/119/ecceb4d...benfrancis:be02dc0.html" title="Last updated on Jul 27, 2022, 4:16 PM UTC (be02dc0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3587,13 +3587,18 @@
 	  <span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeevent-6d">
             The <code>id</code> field SHOULD be set to a unique identifier for 
-            the event, for use when re-establishing a dropped connection.</span>  
-            See below.
+            the event, for use when re-establishing a dropped connection (see 
+            below).</span>
+    <span class="rfc2119-assertion" 
+            id="http-sse-profile-protocol-binding-subscribeevent-6e">
+            It is RECOMMENDED that the identifier is a timestamp 
+            representing the time at which the event ocurred, in the 
+            &quot;date-time&quot; format specified by [[RFC3339]].</span>
           </p>
           <pre class="example">
             event: overheated\n
             data: 90\n
-            id: 12345\n\n
+            id: 2021-11-16T16:53:50.817Z\n\n
           </pre>
 	  <p><span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeevent-7a">
@@ -3748,13 +3753,18 @@
 	  <span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeallevents-4d">
             The <code>id</code> field SHOULD be set to a unique identifier for 
-            the event, for use when re-establishing a dropped connection.</span>
-         See below.
+            the event, for use when re-establishing a dropped connection (see 
+            below).</span>
+    <span class="rfc2119-assertion" 
+            id="http-sse-profile-protocol-binding-subscribeallevents-4e">
+            It is RECOMMENDED that the identifier is a timestamp 
+            representing the time at which the event ocurred, in the 
+            &quot;date-time&quot; format specified by [[RFC3339]].</span>
       </p>
           <pre class="example">
             event: overheated\n
             data: 90\n
-            id: 12345\n\n
+            id: 2021-11-16T16:53:50.817Z\n\n
           </pre>
 	  <p><span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeallevents-5a">


index.html
```diff
@@ -2374,12 +2374,14 @@ <h5><code>subscribeevent</code></h5>
             <code>EventAffordance</code> and MUST be serialized in JSON.
             The <code>id</code> field SHOULD be set to a unique identifier for 
             the event, for use when re-establishing a dropped connection (see 
-            below).
+            below). It is RECOMMENDED that the identifier is a timestamp
```
mlagally, line 2377: Recommending ISO 8601 is not enough  to guarantee interop.

mlagally: I think we should define a structured event payload that contains the timestamp and id as members. 
We will do other event bindings, and this protocol specific encoding will prevent interop.

mlagally (changes requested):

index.html
benfrancis: @mlagally wrote:
> Recommending ISO 8601 is not enough to guarantee interop.

Can you explain why? I have a feeling that the potential ambiguities you are concerned about (e.g. 24 vs. 00 and the Z separator) are fixed in the latest version.

> I think we should define a structured event payload that contains the timestamp and id as members.
We will do other event bindings, and this protocol specific encoding will prevent interop.

Again, protocol bindings are meant to be protocol specific, that's the whole point. See https://github.com/w3c/wot-profile/issues/126#issuecomment-971507879

benfrancis: Potential alternative reference: [RFC3339](https://www.rfc-editor.org/rfc/rfc3339)

benfrancis:
> Ben to resolve merge conflict and propose reference change then re-submit for review

I've re-based against main and made the RFC3339 reference consistent with the observeproperty bindings.

(This may need rebasing again if #249 lands first since the indentation is currently a bit of a mess.)

index.html
```diff
@@ -3587,13 +3587,18 @@ <h5><code>subscribeevent</code></h5>
 	  <span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeevent-6d">
             The <code>id</code> field SHOULD be set to a unique identifier for
```
mlagally, line 3589: Is it permitted to have two events with the same id and different notification content?

index.html
```diff
@@ -3587,13 +3587,18 @@ <h5><code>subscribeevent</code></h5>
...
             The <code>id</code> field SHOULD be set to a unique identifier for 
-            the event, for use when re-establishing a dropped connection.</span>  
-            See below.
+            the event, for use when re-establishing a dropped connection (see 
+            below).</span>
+    <span class="rfc2119-assertion" 
+            id="http-sse-profile-protocol-binding-subscribeevent-6e">
+            It is RECOMMENDED that the identifier is a timestamp 
+            representing the time at which the event ocurred, in the 
+            &quot;date-time&quot; format specified by [[RFC3339]].</span>
```
mlagally, line 3596: I like this recommendation. Should we also provide a fallback mechanism, e.g. using numbers that get incremented when no timestamp is available?

index.html
```diff
@@ -3748,13 +3753,18 @@ <h5><code>subscribeallevents</code></h5>
 	  <span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeallevents-4d">
             The <code>id</code> field SHOULD be set to a unique identifier for 
-            the event, for use when re-establishing a dropped connection.</span>
-         See below.
+            the event, for use when re-establishing a dropped connection (see
```
mlagally, line 3756: see previous comments.

index.html
benfrancis: No, that's what "unique identifier" means :)

Event affordances emit multiple events with the same event name but with a unique identifier and potentially different notification content in each.

index.html
benfrancis: I'm not opposed to defining a fallback, but what if the device doesn't have non-volatile storage and can't keep track of an incrementing number across reboots? How many layers of fallback should we define..?

mlagally:
Arch call on Aug 3rd.: merge

benfrancis:
I've filed #252 for a follow-up discussion on a fallback ID mechanism for devices without a clock.

PR #121, opened 2021-11-08
Add Discovery section to HTTP Baseline Profile - closes #112

### LINKED ISSUE #112: Add Discovery section to Core Profile
Previously suggested in https://github.com/w3c/wot-profile/issues/73#issuecomment-804252286 and https://github.com/w3c/wot-profile/pull/78.

I propose adding a subsection to the WoT Core Profile section called "Discovery" which defines the discovery mechanisms (if any) which Core Profile conformant Consumers MUST support.

From https://github.com/w3c/wot-profile/pull/78#discussion_r742900701:

> Of the introduction mechanisms defined by the WoT Discovery specification, I think [DNS-based Service Discovery](https://w3c.github.io/wot-discovery/#introduction-dns-sd) could be a candidate for a mandatory discovery mechanism in the Core Profile, but that may be asking too much.
> 
> My current thinking is that  [Direct](https://w3c.github.io/wot-discovery/#introduction-direct) discovery could be made the mandatory discovery mechanism by simply requiring that a Thing Description conforming to the Core Profile must be resolvable through an HTTP URL which can be manually typed or copy & pasted in order to locate the Thing Description.

benfrancis:
This PR adds a Discovery section to the Core Profile as suggested in #112.

I propose that the only mandatory discovery mechanism for the Core Profile should be Direct discovery, simply meaning that the Thing's Thing Description must be retrievable from an HTTP URL.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/121.html" title="Last updated on Nov 17, 2021, 4:20 PM UTC (5f83d3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/121/316ba88...benfrancis:5f83d3d.html" title="Last updated on Nov 17, 2021, 4:20 PM UTC (5f83d3d)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/121.html" title="Last updated on Jul 27, 2022, 4:52 PM UTC (321d827)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/121/ecceb4d...benfrancis:321d827.html" title="Last updated on Jul 27, 2022, 4:52 PM UTC (321d827)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2970,6 +2970,21 @@
         WoT Security Best Practices</a> for implementation advice.
       </p>
     </section>
+
+    <!-- Discovery -->
+    <section>
+      <h2 id="http-baseline-profile-discovery">Discovery</h2>
+      <p class="rfc2119-assertion" id="http-baseline-profile-discovery-1">
+        In order to conform with the HTTP Baseline Profile, a Web Thing's Thing 
+        Description [[wot-thing-description]] MUST be retrievable from a 
+        <a href="https://w3c.github.io/wot-architecture/#dfn-wot-thing-description-server">
+        Thing Description Server</a> [[wot-architecture11]] using an HTTP 
+        [[HTTP11]] URL provided by a 
+        <a href="https://w3c.github.io/wot-discovery/#introduction-mech">
+        Direct Introduction Mechanism</a> [[wot-discovery]].
+      </p>
+    </section>
+  </section>
   </section>
 
 


mlagally:
I think requiring the directory API is not what we want to achieve here.
I support the self-descriptive approach and suggest that the security TF considers to isolate the well-know URI
discovery, which then could be adopted by the profile.

benfrancis:
@mlagally wrote:
> I think requiring the directory API is not what we want to achieve here.

I agree that wasn't the intention, and that would be a separate discussion.

> I support the self-descriptive approach and suggest that the security TF considers to isolate the well-know URI
> discovery, which then could be adopted by the profile.

Putting the issue of the Directory Service API aside, I still don't think it's possible to make the [Well-Known URI discovery mechanism](https://w3c.github.io/wot-discovery/#introduction-well-known) mandatory for all Web Things conforming to the Core Profile. You can only have one well-known URI per origin, so it doesn't work for cases where multiple Thing Descriptions are hosted from the same origin.

I don't think this is something that the discovery task force (I assume you meant them, and not the security task force?) can fix, because it's inherent in the way well-known URIs work (see [RFC8615](https://datatracker.ietf.org/doc/html/rfc8615)).

My recommendation is therefore that only the [Direct](https://w3c.github.io/wot-discovery/) discovery mechanism (i.e. that a Thing Description must be retrievable from an HTTP URL) should be mandatory for the Core Profile, so we should land this PR as it is.

mmccool:
For reasons other than those cited (which I think are all fixable) I would not recommend making .well-known mandatory since it's only useful for self-description, and making self-description mandatory has privacy implications.  Note however it is mostly for the case where you don't have the real URL, just e.g. an IP address.

mlagally:
@mmccool 

> For reasons other than those cited (which I think are all fixable) I would not recommend making .well-known mandatory since it's only useful for self-description, and making self-description mandatory has privacy* implications. Note however it is mostly for the case where you don't have the real URL, just e.g. an IP address.

I think the privacy problem can be solved even with a mandatory self description mechanism.
There are two cases:
- If the caller is trusted by the device (or no privacy sensitive information is kept in the device) the well known URL reorts full data.
- in the other cases, it just reports the kind of information that can be freely exposed to anyone. In some cases this may be very limited content or even no content at all.

With these thoughts, I suggest to make the well known URL mechanism mandatory.

mlagally:
As of today, the current discovery specification is in flux and may create significant additional implementation burden on the consumer. 
See https://github.com/w3c/wot-discovery/issues/272 for details.

I suggest we hold off with this MR until the discovery spec is consolidated. 

After that we should consider whether the constraints that are defined in the discovery specification are causing significant implementation burden.

mlagally (commented):
Suggest we wait until the discovery spec is more stable.

benfrancis:
Proposal based on https://github.com/w3c/wot-discovery/issues/263#issuecomment-1099402892...

How about we say that for the HTTP Baseline Profile:
1. Conformant Things MUST have a Thing Description which can be fetched from an HTTP URL
2. Conformant Consumers MUST be able to fetch a Thing Description from an HTTP URL

In other words, Things and Consumers must support the [Direct introduction mechanism](https://w3c.github.io/wot-discovery/#introduction-direct) and [Self-description exploration mechanism](https://w3c.github.io/wot-discovery/#exploration-self), but don't necessarily have to implement Thing Directories and Thing Links.

mlagally:
Arch call on 11.5.:
PR has merge conflicts. 
It puts a mandatory requirement on system behavior that is too narrow, i.e. prevents out of band communication of a TD (e.g. via email)
We could add a link to the TD with a URL where the TD is hosted, this could also be a self descriptive URL.

benfrancis:
Yes the PR needs some work, but I was hoping to agree on the general requirement first.

> It puts a mandatory requirement on system behavior that is too narrow, i.e. prevents out of band communication of a TD (e.g. via email)

It doesn't prevent the TD being shared by email, it just requires that TDs which conform to the HTTP Baseline profile can be fetched over HTTP. That means that any conforming consumer will be able to discover any conforming Thing using the [Direct introduction mechanism](https://w3c.github.io/wot-discovery/#introduction-direct). If no discovery mechanisms are mandatory then the TD may not be discoverable at all by conforming consumers, and therefore not interoperable.

> We could add a link to the TD with a URL where the TD is hosted, this could also be a self descriptive URL.

I see now harm in a self link in the TD, but how does that help with discovery?

benfrancis:
It's currently [looking like](https://github.com/w3c/wot-discovery/pull/313) support for "[security bootstrapping](https://github.com/w3c/wot-discovery/issues/135)" (aka "Thing Description authentication") may be made optional in the WoT Discovery specification. This may mean that out-of-the box interoperability between WoT Consumers and WoT Directories (or in fact self-describing Things) is not possible for Thing Descriptions which require authentication.

If this is the case, I think it further strengthens the argument to add a Discovery section to the HTTP Profile which adds additional constraints on top of what is described in the WoT Discovery specification. Specifically:
1. Conformant Things MUST have a Thing Description which can be fetched from an HTTP URL
2. Conformant Consumers MUST be able to fetch a Thing Description from an HTTP URL
3. Conformant Things, Directories and Consumers MUST support a finite list of authentication mechanisms for accessing Thing Descriptions over HTTP

Without these constraints it's likely that (where TD authentication is required) only WoT Consumers and WoT Directories (and Things) from the same vendor will work together, since Consumers will require custom code for each Directory implementation they consume.

See also: https://github.com/w3c/wot-discovery/pull/313#issuecomment-1141235764.

mmccool:
I am not against some additional constraints on Discovery in the http-baseline-profile but want to make some observations:
- Support for Discovery itself is optional (SHOULD) in the WoT Architecture document.  Making it mandatory (subject to some constraints) would help interoperability, certainly.
- The "Direct" introduction method does not deliver TDs, it is an "Introduction mechanism" and is just a means of providing a URL that could also point at a Directory (for example).  The PR as written seems to combine the concepts of the Direct Introduction and a TD Server (which actually is the service that returns a TD when given a URL from an Introduction mechanism).  Also "Direct" is kind of a catch-all for "other" sources of URLs and is not really the right thing for improving interop, since we still don't know where the URL will come from.  What Discovery does require is that at least one introduction is supported.  
- I think it is reasonable to insist that Things conforming to the profile support both the well-known and direct introduction mechanisms.  Earlier I had concerns about privacy with well-known, but that can be mitigated by supporting authentication (perhaps with security bootstrapping).  In practice what you would want to do is use security bootstrapping just for the initial connection of a Thing to a hub, and then (once you have established better security, e.g. installing PSK) disabling the security bootstrapping.  In other words, a onboarding process.  Really this process should be a whole other deliverable (and I plan to propose we work on such a deliverable in our next charter) but we could perhaps lay the groundwork for it.  Once you have set up a "group key" for accessing TDs at least, then the .well-known URL is still useful for all the devices using that key. 
- Two things we could do in addition for profiles are: 1. providing a common way to disable security bootstrapping 2. providing a common way to establish a group key for accessing TDs via .well-known.  Steps 1 and 2 could be combined in a single standardized action.   It might be a good idea to also define a common "offboarding" action that removes these keys and returns it to its original state, and a side-effect of this should also be to delete any other configuration data.

index.html
```diff
@@ -2608,8 +2608,26 @@ <h4>Error Responses</h4>
...
       </section>
     </section>
-  </section>
 
+    <!-- Discovery -->
+    <section>
+      <h2 id="discovery">Discovery</h2>
+      <p class="rfc2119-assertion" id="core-profile-discovery-1">
+        In order to a conform with the Core Profile, a Web Thing's Thing
+        Description [[wot-thing-description]] MUST be retrievable from a URL
```
mmccool, line 2617: With the new terminology, what you want to say here is the Thing Description for a Thing must be available from a TD Server, given a URL provided with (at least) Direct introduction (which just means the user needs to provide the URL, somehow).

See my other comments.  Given we are talking greenfield devices, it might be best after all if we require .well-known and self-description, as long as we can resolve the privacy/security issues.

index.html
```diff
@@ -2608,8 +2608,26 @@ <h4>Error Responses</h4>
...
+        [[URL]] using the HTTP protocol [[HTTP11]], to enable its discovery
+        using a 
+        <a href="https://w3c.github.io/wot-discovery/#introduction-direct">
+        Direct Introduction Mechanism</a> [[wot-discovery]].
+      </p>
+      <p class="rfc2119-assertion" id="core-profile-discovery-2">
+        In order to a conform with the Core Profile, a Consumer MUST be
```
mmccool, line 2624: maybe "retrieve a TD from a TD Server" using the new terminology in WoT Discovery.

mmccool (commented):
I think with a few tweaks, to mention a TD Server that the URL provided by Direct points at (a Direct Introduction itself does not provide a TD, it points at a TD Server that provides it) this could be merged.  But the whole problem does need more thought.

mlagally:
> It's currently [looking like](https://github.com/w3c/wot-discovery/pull/313) support for "[security bootstrapping](https://github.com/w3c/wot-discovery/issues/135)" (aka "Thing Description authentication") may be made optional in the WoT Discovery specification. This may mean that out-of-the box interoperability between WoT Consumers and WoT Directories (or in fact self-describing Things) is not possible for Thing Descriptions which require authentication.

We could (and should) use a self descriptive mechanism, which may require authentication. This would be up to the device manufacturer to decide.
For devices without sensitive TDs, the well known URI returns the full TD. For sensitive devices, if the consumer does not authenticate, the well known URI just returns a crippled TD that only contains info about the supported auth schemes. After successful authentication of a consumer, it would return the full TD.

All OOTBI scenarios would work, if the consumer knows the credentials. 

> 
> If this is the case, I think it further strengthens the argument to add a Discovery section to the HTTP Profile which adds additional constraints on top of what is described in the WoT Discovery specification. Specifically:
> 
> 1. Conformant Things MUST have a Thing Description which can be fetched from an HTTP URL
This requirement seems a bit restrictive and needs some clarification. What is the purpose of this constraint? What if somebody provides a TD over a different protocol? The Thing would be completely usable and all OOTB use cases would work. What if somebody uploads a TD to a server in an HTML form?

> 2. Conformant Consumers MUST be able to fetch a Thing Description from an HTTP URL
Same concerns as on #1.

> 3. Conformant Things, Directories and Consumers MUST support a finite list of authentication mechanisms for accessing Thing Descriptions over HTTP

I'm ok with the finite set of auth mechanisms, we may be even able to define the list. 
However so far the profile does not talk about directories at all and I don't think the profile should mandate them. The deployment scenario / business case for directories needs to be clarified.

> 
> Without this third constraint it's likely that (where TD authentication is required) only WoT Consumers and WoT Directories (and Things) from the same vendor will work together, since Consumers will require custom code for each Directory implementation they consume.
> 
> See also: [w3c/wot-discovery#313 (comment)](https://github.com/w3c/wot-discovery/pull/313#issuecomment-1141235764).

mlagally:
> I am not against some additional constraints on Discovery in the http-baseline-profile but want to make some observations:
> 
> * Support for Discovery itself is optional (SHOULD) in the WoT Architecture document.  Making it mandatory (subject to some constraints) would help interoperability, certainly.
I think the profile should not make directories mandatory, because profile compliant things will work in many use cases that do not require them.

> * I think it is reasonable to insist that Things conforming to the profile support both the well-known and direct introduction mechanisms.  Earlier I had concerns about privacy with well-known, but that can be mitigated by supporting authentication (perhaps with security bootstrapping).  In practice what you would want to do is use security bootstrapping just for the initial connection of a Thing to a hub, and then (once you have established better security, e.g. installing PSK) disabling the security bootstrapping.  In other words, a onboarding process.  Really this process should be a whole other deliverable (and I plan to propose we work on such a deliverable in our next charter) but we could perhaps lay the groundwork for it.  Once you have set up a "group key" for accessing TDs at least, then the .well-known URL is still useful for all the devices using that key.
This is similar to the approach in my previous comment, bnut it needs more thought. Who would provide this "group key"? Can it be revoked/updated when a device is transferred to a new owner?

mlagally:
@benfrancis 
please change "core" to "baseline"

mlagally:
Arch call on July 20th:
It seems to ok to require that TDs have a HTTP URL, wording of the spec needs to be aligned with the discovery spec. 
Proposed way forward: upate language, resolve merge conflict.

benfrancis:
OK, I've reduced this down to the simplest, least contentious form I can think of:

> In order to conform with the HTTP Baseline Profile, a Web Thing's Thing Description [[wot-thing-description](https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/121.html#bib-wot-thing-description)] MUST be retrievable from a[ Thing Description Server](https://w3c.github.io/wot-architecture/#dfn-wot-thing-description-server) [[wot-architecture11](https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/121.html#bib-wot-architecture11)] using an HTTP [[HTTP11](https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/121.html#bib-http11)] URL provided by a[ Direct Introduction Mechanism](https://w3c.github.io/wot-discovery/#introduction-mech) [[wot-discovery](https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/121.html#bib-wot-discovery)].

Hopefully this establishes the most basic requirement of discovery via HTTP.

----

I think two further points require further discussion:
1. Whether to make some form of security bootstrapping mandatory
2. Whether to make the well-known introduction mechanism mandatory

I would like to do the former if possible, but it will require defining a fixed set of authentication mechanisms. I don't think the latter is necessarily a good idea, because of the limitations of only one well-known URL per origin.

benfrancis:
@mlagally wrote:
> For devices without sensitive TDs, the well known URI returns the full TD. For sensitive devices, if the consumer does not authenticate, the well known URI just returns a crippled TD that only contains info about the supported auth schemes. After successful authentication of a consumer, it would return the full TD.

I originally thought this was a good idea too, but we've now established that this doesn't work as a bootstrapping mechanism because the security metadata in a Thing Description refers to the endpoints of interaction affordances, not to the Thing Description resource itself. See https://github.com/w3c/wot-discovery/issues/135.

The WoT Discovery specification now describes another mechanism in [7.1.2 Security Bootstrapping](https://w3c.github.io/wot-discovery/#exploration-secboot) which uses HTTP 401, 302 and 303 response codes (depending on the security mechansim being used). Support for this bootstrapping mechanism is not currently mandatory for either Producers or Consumers, so as things stand only unauthenticated Thing Descriptions (consumed by Consumers which support HTTP and a Direct introduction mechanism) are guaranteed to be discoverable with the above addition to the HTTP Baseline Profile.

mlagally:
@benfrancis Thanks for fixing the merge conflict.
The discovery section targets the baseline profile only, so strictly speaking TDs that conform with the SSE or WebHook profile won't be discoverable. 
I suggest you move the proposed text to chapter 6, which should contain all shared mechanisms and constraints.

index.html
```diff
@@ -2970,6 +2970,21 @@ <h2 id="http-baseline-profile-security">Security</h3>
         WoT Security Best Practices</a> for implementation advice.
       </p>
     </section>
+
+    <!-- Discovery -->
+    <section>
+      <h2 id="http-baseline-profile-discovery">Discovery</h2>
+      <p class="rfc2119-assertion" id="http-baseline-profile-discovery-1">
+        In order to conform with the HTTP Baseline Profile, a Web Thing's Thing
```
mlagally, line 2978: I think this language is problematic, since it mandates the presence of a TD server as part of a deployment of profile compliant devices and I don't understand the rationale behind it.

How does the manufacturer of a thing ensure that the TD is deployed in this way? What if all TDs are stored on an old-fashioned FTP server or are stored in the file system (or hard coded) in the consumer?

I don't see a clear business model for a 3rd party to provide a TD server with a revenue stream that pays for running the server.
I'm afraid that if we enforce that constraint, we might risk adoption of the profile.

egekorkan: I sort of agree with this comment but we don´t have to make a one size fits all profile anyways, just an implementable subset is fine. If we are trying to make something that fits all, lots of other parts of the spec are open for change.

benfrancis: @mlagally wrote:
> I think this language is problematic, since it mandates the presence of a TD server as part of a deployment of profile compliant devices and I don't understand the rationale behind it.

A TD server is just an HTTP server which can serve a single JSON Thing Description. A Web Thing conforming to an HTTP already has to have an HTTP server serving JSON resources, so there's no reason the Thing Description (the most important resource of all) couldn't be served over HTTP too.

> How does the manufacturer of a thing ensure that the TD is deployed in this way?

The obvious way is for the Web Thing to serve its own Thing Description, as described above. But they could host it elsewhere if they really wanted to... I don't think I understand the problem.

> What if all TDs are stored on an old-fashioned FTP server or are stored in the file system (or hard coded) in the consumer?

Then it's not compliant with the profile. I don't know why on earth a manufacturer of a greenfield HTTP-based device would want to do either of those things. Hard-coding TDs in a Consumer is particularly counterproductive as far as interoperability is concerned.

> I don't see a clear business model for a 3rd party to provide a TD server with a revenue stream that pays for running the server.

They don't need to. The Web Thing can serve the TD itself.

> I'm afraid that if we enforce that constraint, we might risk adoption of the profile.

I'm afraid that if Thing Descriptions don't have even one single basic mechanism for discovering them, it will risk adoption of the profile. Do you have another proposed mechanism other than the HTTP profile requiring the TD be served over HTTP?

I'm really surprised that having the HTTP Profile require that Thing Descriptions are served over HTTP is in any way controversial. I thought we already agreed that in the last meeting:

> Arch call on July 20th:
> It seems to ok to require that TDs have a HTTP URL, wording of the spec needs to be aligned with the discovery spec.
> Proposed way forward: upate language, resolve merge conflict.

benfrancis:
@mlagally wrote:
> I suggest you move the proposed text to chapter 6, which should contain all shared mechanisms and constraints.

OK, let's do that in #214.

mlagally:
Arch call on Aug 3rd: agree to merge

PR #124, opened 2021-11-09
Move and reword Security section

benfrancis:
As originally proposed in https://github.com/w3c/wot-profile/issues/73#issuecomment-804252286 and https://github.com/w3c/wot-profile/pull/78 this PR moves the Security section out of the WoT Core Data Model section into its own section within Core Profile.

I've also refined the wording of the normative statements and added references to the WoT Thing Description and WoT Security Security Best Practices documents.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/124.html" title="Last updated on Nov 25, 2021, 9:26 AM UTC (0fc1a0a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/124/316ba88...benfrancis:0fc1a0a.html" title="Last updated on Nov 25, 2021, 9:26 AM UTC (0fc1a0a)">Diff</a>


benfrancis:
@mlagally When I split the large PR into six separate PRs, the reason that I kept these changes together is that the one change builds on the other, which means that two separate PRs would cause a merge conflict.

I've now split this PR into two separate commits so you can see the diffs separately, but I'm not sure it actually makes much difference.

Issue #114, opened 2021-11-05
Add Links section to Core Profile


benfrancis:
Previously suggested in https://github.com/w3c/wot-profile/issues/73#issuecomment-804252286 and https://github.com/w3c/wot-profile/pull/78.

I have previously proposed adding a subsection to the WoT Core Profile section called "Links" which defines types of link relations which Core Profile conformant Consumers MUST support.

However, looking through the list of "[best practice relation type](https://w3c.github.io/wot-thing-description/#link)"s in the Thing Description specification I'm not sure there are any which actually make sense as being mandatory.

I had previously suggested that link `rel=alternate` with `type=text/html` could be used to denote an HTML user user interface in the web thing (we do this in WebThings), but in retrospect I'm not sure it's really possible to make this mandatory since supporting it assumes that a Consumer can render a web page to the user, which may not be the case for all consumers.

Are there any link relation types for which support should be mandatory in the Core Profile, or shall we just leave this out?

mlagally:
I think we should include links that allow composition and aggregation of things to a larger structure to the profile.
There are names in the IANA registry that would be appropriate for these.

benfrancis:
@mlagally wrote:
> I think we should include links that allow composition and aggregation of things to a larger structure to the profile.

This seems like quite an advanced feature for a core profile. Are there any existing implementations which provide implementation experience of something like this? Is this something you have plans to implement yourself?

See also: https://w3c.github.io/wot-thing-description/#thing-model-composition

mlagally:
The profile includes a "Links" section, so we should close this issue. 
Conversations about individual requirements in the "Links" section are happening in separate issues.

Issue #29, opened 2020-08-27
Make units mandatory?


mlagally:
Arch call on 27.8.:
This is important to enable interop, need to also select a unit scheme, e.g. qudt.
A unit scheme that is widely adopted is higly preferable.

OneDM took SenML units, aligning could be useful, this is simpler.
OneDM intends to align with IETF, aligning is pragmatic to enhance interop across standards.

Qudt implementation could be very costly.

Proposed Spec Language:
You must use SenML units if the unit exists there. You may only use an extension point if the unit does not exist in SenML. In the case of an extension the use of QUDT is (mandatory|recommended). 

SenML: https://tools.ietf.org/html/rfc8428
QUDT Ontologies Overview: http://www.qudt.org/pages/QUDToverviewPage.html

mlagally:
for units: https://genivi.github.io/vehicle_signal_specification/rule_set/data_entry/data_unit_types/

mlagally:
Arch call on 3.9.
Some domains have very specific units, it may not be ideal to force specific metric systems.
We should adopt well known unit ontologies, it may depend on specific use cases. 

Finite set of unit extensions and a standardized prefix to unambiguously identify the unit.
Also require a way to specify a version and a fixed context string that can be statically parsed.

benfrancis:
I don't agree that units should be mandatory for every property (taking the title literally) as that doesn't always make sense.

I do agree that requiring that Consumers conforming to the Core Profile support a particular ontology for units would be a good idea, since that would aid interoperability. The challenge will be in picking one.

In WebThings we [use](https://webthings.io/schemas/) the long written form of SI units (e.g. hectopascal, percent, kelvin, ampere, micrograms per cubic metre, hertz, watt, degree celsius, volt, second).

Is ISO 80000-3 referenceable? See:
-  https://en.wikipedia.org/wiki/International_System_of_Units
- https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf/2d2b50bf-f2b4-9661-f402-5f9d66e4b507?version=1.10&download=true
- https://www.nist.gov/pml/special-publication-330/sp-330-section-2#2.3

egekorkan:
Comments before the previous one seem to suggest allowing but restricting to multiple ontologies. It is not possible to satisfy everyone, that is why the base TD spec is and will be completely open about it. Given that profile is for a subset of cases, I don't think that we should lose time trying to find something that satisfies all and pick one or not pick anything at all and be open.

[...]

Issue #60, opened 2020-12-18
Allow multiple forms for interaction affordances in Core Profile


benfrancis:
Section [5.1.4.2 Additional Constraints](https://www.w3.org/TR/2020/WD-wot-profile-20201124/#additional-constraints) of the First Public Working Draft says:
> The Array of Form of each property MUST contain only a single endpoint for each operation readproperty, writeproperty, observeproperty, unobserveproperty.

This constraint has the effect of preventing a WoT producer from supporting multiple protocols for its interaction affordances whilst conforming with the Core Profile. This seems like an unnecessary constraint as it doesn't help interoperability and may even prevent a WoT producer from conforming to multiple profiles.

I propose that this constraint should be dropped so that it's possible for a WoT device to be conformant with the Core Profile (e.g. using the default HTTP protocol binding) whilst also exposing endpoints using other protocols (e.g. WebSockets).

This would also make it easier to modify brownfield devices to make them conformant with the Core Profile, by allowing them to add new endpoints conforming to the Core Profile whilst continuing to expose other endpoints using legacy protocols and protocol bindings.

See also: https://lists.w3.org/Archives/Public/public-wot-wg/2020Dec/0025.html

mlagally:
Arch call on 21.1.:
This has been written with multple forms for the same protocol in mind.
Need to revisit and could constrain to "only one form per protocol."

mlagally:
There's an overlap of this issue with #141

benfrancis:
Note: As I understand it this issue is about removing the current constraint which forbids multiple forms per interaction affordance and #141 is about how a Consumer should choose between multiple forms if they exist.

relu91:
> Need to revisit and could constrain to "only one form per protocol."

As @sebastiankb mentioned in https://github.com/w3c/wot-profile/pull/125#issuecomment-966199892: 
> I also see a problematic conflict that forms should only have one entry. Servients can have typically multipile IP addresses such a local IP and global IP address. Sometimes you get also variations for IPv4 and IPv6. note-wot is the best example of how multiple form entries are possible even for a single protocol like http. Depending on where the consumers host, they can choose a local or global IP address for interaction.

Having multiple networks interfaces is pretty common in our set-ups. Forcing to have just one form per-protocol means that users will need to spawn N WebThings (most of which will be just proxies) with N equal to the number of the networking interfaces. I'd rather have a well-defined algorithm to choose a form as requested in #141.

benfrancis:
[...]

Issue #59, opened 2020-12-18
Consider splitting Core Profile into Core Profile and Constrained Profile


benfrancis:
The [First Public Working Draft](https://www.w3.org/TR/2020/WD-wot-profile-20201124) says the goals of the WoT Core Profile are:
> - guarantee interoperability among all implementations of the profile.
> - limit the implementation complexity for resource constrained devices.
> - ensure that a thing description is also useful for a human reader.

Looking at this set of goals, I would suggest that the first goal has not yet been met by this draft of the specification and the second and third goals are causing conflicting requirements.

As mentioned in #24, it is not possible to guarantee interoperability between all implementations of a profile unless all implementations are required to support at least one protocol (and serialisation format). If a WoT producer only supports CoAP + CBOR and a WoT consumer only supports HTTP + JSON then the fact that they both share the same abstract data model is irrelevant, they will be not be able to communicate with each other. Not requiring implementations of the Core Profile to support at least one protocol therefore does not meet the requirement of out of the box interoperability as defined in the Abstract of the specification.

I would also suggest that the goal of making thing descriptions human readable and the goal of limiting implementation complexity for constrained devices cause too many conflicting requirements. On the one hand the Core Profile limits the length of strings in an attempt to reduce resource needs for resource constrained devices, but on the other hand it adds a large number of mandatory fields and human readable strings which add to those resource needs.

I propose that both of these problems could be solved by splitting the Core Profile into two separate profiles:
1. **Core Profile** - requiring support for HTTP + JSON, with a focus on human readability (replacing the current [Protocol Binding based on HTTP](https://w3c.github.io/wot-thing-description/#http-binding-assertions) section of the Thing Description specification, fixing #11)
2. **Constrained Profile** - requiring support for CoAP + CBOR, with a focus on reducing implementation complexity for resource constrained devices.

mlagally:
Arch call on 21.1.:
- profile spec is prescriptive vs. TD spec, which is descriptive
- we cannot add prescriptive constraints to the TD spec without breaking compatibility
- we will focus on the core data model for the next couple of calls
- agreed on limitation of depth of max 5 in the data model
- profile spec has conformance guarantees and rules that can be validated
- an implementable profile has at least one protocol binding
- we need to define a prescriptive protocol binding for HTTP (somewhere)
- constraints on protocol bindings for MQTT / CoAP + CBOR binding should be considered after we conclude on HTTP

benfrancis:
[...]

PR #125, opened 2021-11-09
Remove WoT Core Data Model and update Core Profile introduction - closes #10

### LINKED ISSUE #10: Reduce or remove the constraints in section 7.2 Baseline Information Model

benfrancis:
This PR proposes removing the WoT Core Data Model section as discussed in #73.

In summary, whilst I don't disagree in principle with the idea of a Data Model section of the Core Profile, this should only include constraints which:
1) Contribute to the agreed new focused goal of the Core Profile, which is interoperability (not the previous additional goals around human readability and resource constrained devices)
2) Can not be part of the [TD information model](https://w3c.github.io/wot-thing-description/#sec-vocabulary-definition) section of the Thing Description specification, which should be considered the real core data model

After extensive review (detailed in https://github.com/w3c/wot-profile/pull/93#pullrequestreview-717982575) my conclusion is that almost none of the constraints currently proposed in the WoT Data Model section meet both of these criteria, and therefore I am proposing the removal of the whole section, with the exception of the [Links](https://github.com/w3c/wot-profile/issues/114) and [Security](https://github.com/w3c/wot-profile/pull/124) sections which I have proposed moving out into their own sections. There's also an [ongoing discussion](https://github.com/w3c/wot-profile/issues/29) about units.

Note that many of the current set of data model constraints are aimed at the previous goals of human readability and reducing resource consumption for resource constrained devices, which we have now agreed work against each other and therefore should not be goals of the Core Profile.

As [requested](https://github.com/w3c/wot-profile/pull/78#issuecomment-960685134), I have split this change out from #78, which also included other parts of the reorganisation proposed in #73 which have now been separated out into their own PRs. With separate PRs it is harder to see the overall effect of that proposal, and merging all of the individual PRs requires resolving merge conflicts. To help with this, you can see a rendered view of the combined effect of my current nine open pull requests here https://benfrancis.github.io/wot-profile/.

I understand that @mlagally does not agree with this part of the proposal, so if we can not reach a consensus then I would like to propose that in line with the [Working Group's Decision Policy](https://www.w3.org/2020/01/wot-wg-charter.html#decisions), the Chairs (@mmccool and @sebastiankb) consider calling a group vote to resolve the conflict.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/125.html" title="Last updated on Nov 17, 2021, 4:39 PM UTC (d138687)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/125/d51d816...benfrancis:d138687.html" title="Last updated on Nov 17, 2021, 4:39 PM UTC (d138687)">Diff</a>


sebastiankb:
The removing of the whole chapter seems a bit harsh to me. 

I took another look at the data model section, and there are definitely also some parts that I don't see the motivation behind forcing them into generic profile design. I'm agree with @benfrancis that many of them addresses human readability like having title and description everywhere.

I also see a problematic conflict that forms should only have one entry. Servients can have typically multipile IP addresses such a local IP and global IP address. Sometimes you get also variations for IPv4 and IPv6. note-wot is the best example of how multiple form entries are possible even for a single protocol like http. Depending on where the consumer hosts, they can choose a local or global IP address for interaction.  

I do not want to go in detail in each aspect. I can understand @mlagally view that he wants to have a certain set of information that a consumer can expect. But the question is who is the potential consumer and I think this question is not well covered yet in the document. 

I'm loud thinking how we can find a compromise here: Since there is a lot of discussion about this data model topic in the past maybe we can mitigate some aspects e.g., by saying that they are highly RECOMMENDED. Or, that having `title` mandatory for the interactions affordances and for the data schema it is recommended etc.

I will join today's meeting where we can further discuss this topic.

benfrancis:
> The removing of the whole chapter seems a bit harsh to me.

OK, which specific normative statements would you keep, and why? I have proposed keeping the Security and Links sections, but making them separate sections, since they do contribute to interoperability.

> maybe we can mitigate some aspects e.g., by saying that they are highly RECOMMENDED. Or, that having `title` mandatory for the interactions affordances and for the data schema it is recommended etc.

How does making `title` mandatory contribute to the goal of the Profile specification, which is to enable out-of-the-box interoperability? What's the reason for `title` to be mandatory in the Core Profile, but not mandatory for all Thing Descriptions? If something is only recommended, then why is it not recommended for all Thing Descriptions?

I'm happy to discuss this further on the call today, but in my view we shouldn't just keep text because it has already been written and therefore it's "harsh" to remove it, it must be contributing to the goals of the specification in some way. If it isn't then it should be removed (regardless of who wrote it), and removing text should be seen as a positive thing since it will reduce implementation complexity. It seems totally reasonable to me to remove large chunks of text since we drastically reduced the goals and scope of the specification, it's not personal and nobody should be offended by that.

sebastiankb:
>    OK, which specific normative statements would you keep, and why? I have proposed keeping the Security and Links sections, but making them separate sections, since they do contribute to interoperability.

One suggestion would be to keep all sections, but make them weaker by making them a recommendation rather than a MUST.  

> How does making title mandatory contribute to the goal of the Profile specification, which is to enable out-of-the-box interoperability? 

You are right, this do increase out-of-the-box interoperability. 

> What's the reason for title to be mandatory in the Core Profile, but not mandatory for all Thing Descriptions? If something is only recommended, then why is it not recommended for all Thing Descriptions?

Well, that's the idea of the profile, to follow a convention that some generic definitions of TD like protocol usage or optional terms should be prescribed here (e.g., the profile prescribe to use HTTP protocol). TD itself is independent of profile definitions. In my opinion, recommendations should be prescribed mainly for specific use cases or profile documents.

mlagally:
I think we are circling around these common constraints since a while and I provided explanations and justifications since several months.
The purpose of this section is to describe common constraints that are valid for things *independent* from the protocol binding. 
Time formats, units, location information and the like are not related to a specific protocol binding. Why should a CoAP binding use different time formats? 
A different way to represent units?
The same commonality is valid for certain metadata fields. 
In the TD almost all metadata fields are optional, so how can a consumer build a reasonable UI without a lot of "out of band" knowledge? What are the minimum mandatory metadata fields to enable this? 
If a consumer cannot rely on these values being present - how can he interoperate with the thing at all? I have seen TDs in plug fests that basically had properties A1..A10 and actions A1..A10 and no further content. Yes, you can read, write and call those, but what does a UI do?

benfrancis:
@sebastiankb wrote:
> You are right, this do not increase out-of-the-box interoperability.

For me, this is the key issue. If the constraints don't contribute to the goal of the specification, then why are they there?

@mlagally wrote:
> I think we are circling around these common constraints since a while and I provided explanations and justifications since several months.

I realise that you feel that you have already responded, and I'm sorry that this discussion is still dragging out. But from my perspective you have only provided vague high level answers which don't really explain how these constraints contribute to the interoperability goal of the specification, or address the problem that many of the constraints conflict with each other. You have still not responded to any of my questions and feedback on specific normative statements in https://github.com/w3c/wot-profile/pull/93#pullrequestreview-717982575.

To recap, the Introduction to the specification currently says that the WoT Core Profile was defined with the following main goals:

>> - guarantee interoperability among all implementations of the profile.
>> - limit the implementation complexity for resource constrained devices.
>> - ensure that a thing description is also useful for a human reader.

It is my understanding that many months ago a decision was made to [change the goals and scope](https://github.com/w3c/wot-profile/issues/73) of the specification focus only on the first goal and drop the latter two, in part because they were found to conflict with each other. All these months later the specification has still not been updated to reflect this decision. Many of the data model constraints were written to address the latter two goals, so what I'm trying to do is remove parts of the specification which were focused on those goals (and often conflicted with each other), and keep only the parts which specifically focus on the goal of interoperability.

> The purpose of this section is to describe common constraints that are valid for things _independent_ from the protocol binding.

This is a bit confusing to me. The WoT Core Data Model is currently part of the Core Profile. It is my expectation that if we add a new protocol binding then it would likely be part of a new profile (e.g. a Constrained Profile with a CoAP+CBOR binding). I would expect that a Constrained Profile would need a different set of constraints to the Core Profile, since it is aimed specifically at constrained devices, whereas the Core Profile uses HTTP and will likely be used by devices with more resources.

In order to resolve the issue of conflicting goals, part of my proposal in #73 was to split the profiles into a Core Profile (using HTTP+JSON) and a Constrained Profile (using CoAP+CBOR), each with their own set of constraints.

Since this first version of the Core Profile will only define one profile (the Core Profile), then perhaps a first step would be to remove any constraints which are aimed at supporting resource constrained devices, and save those for a future Constrained Profile instead.

> Why should a CoAP binding use different time formats?

I don't necessarily think that it should, but currently we are only defining an HTTP binding. The time format constraint is a recent addition to the data model which I wouldn't be opposed to keeping as part of the Core Profile.

> A different way to represent units?

Regarding units, as I said in https://github.com/w3c/wot-profile/pull/93#pullrequestreview-717982575, _"if this section constrained conformant Web Things to a particular ontology or set of ontologies for units then it could actually be useful for interoperability."_ There are discussions around that in #29 but it's looking like it may not be possible. It's possible that you might pick a different unit ontology for constrained devices (e.g. which has a more efficient representation, or a more constrained set of units), but this isn't something that I've looked into.

> In the TD almost all metadata fields are optional, so how can a consumer build a reasonable UI without a lot of "out of band" knowledge? What are the minimum mandatory metadata fields to enable this?
> If a consumer cannot rely on these values being present - how can he interoperate with the thing at all? I have seen TDs in plug fests that basically had properties A1..A10 and actions A1..A10 and no further content. Yes, you can read, write and call those, but what does a UI do?

OK, so this is addressing the goal of human readability, which is separate from the goals of interoperability or limiting the implementation complexity for resource constrained devices.

It was my understanding that we had agreed that human readability was no longer a goal of the Core Profile. If you think that this should still be a goal, then we need to re-open that discussion. If the group decides that we should bring back that goal, then we could discuss which of the human readability constraints to keep in the Core Profile.

Finally, if there are constraints which really would be applicable to _all_ profiles and protocol bindings, then my suggestion would be to make them part of the Thing Description specification instead, which should be considered the real core data model.

In conclusion, if you still feel that a Core Data Model is needed as part of the Core Profile, my suggestion would be to:
1. Remove any constraints focused on reducing implementation complexity for constrained devices, and save those for a future Constrained Profile
2. Discuss whether human readability is still a goal of the Core Profile, and if so then discuss which constraints to keep (e.g. `title` being mandatory for `InteractionAffordance`s, and `description` being recommended). If not, then remove all constraints which are focused only on human readability.
3. Evaluate each of the remaining constraints (taking into account my detailed feedback in https://github.com/w3c/wot-profile/pull/93#pullrequestreview-717982575) to determine whether they contribute to interoperability, whether they are really necessary, and whether they would be better suited to recommendations in the Thing Description specification instead.

Currently the only constraints that I personally feel benefit interoperabilty are:
1. A constrained set of security schemes (which I proposed moving to a separate security section)
2. A constrained set of link relations (which I proposed moving to a separate links section)
3. A constrained set of discovery mechanisms (which I proposed moving to a separate discovery section
4. A very limited set of data constraints on things like time formats and units which might otherwise prevent interoperability

In the next charter period I think we should work on a Constrained Profile, and consider re-introducing some of the data model constraints that help with the goal of limiting implementation complexity for constrained devices. At that point we might also want to consider splitting out a set of "[common constraints](https://github.com/w3c/wot-profile/pull/131)" which apply to all profiles, but for now I think we should just stick to defining the Core Profile, which has a clear set of requirements.

mryuichi:
I don't think it's necessary to delete all the items in the core data model session. The reason is as follows.

Out-of-the-box interoperability makes It easier for consumers to connect to their devices. Unfortunately human-readable has little to do with this. However, to set up an IoT system, not only immediate use but also subsequent maintenance is an important factor. After a while, it's easy to lose track of what the connected device is, what features it has, and who manages it. It is an important requirement that the information required for maintenance can be easily obtained as an impact when out-of-the-box interoperability is realized.
The human-readable items described in the core data model are important information that helps the maintenance worker to understand. It is safe to use the profile document to define the items to be described in common even if multi-vendor devices are connected.

Why not expand the interpretation of out-of-the-box interoperability to include not only protocol binding and discovery, but also post-connection maintenance?

Remember that Fujitsu used to connect the smart home in the plugfest? More than 200 devices of 30 kinds such as sensors, home appliances, and batteries are connected to that home. All the devices are ECHONET Lite compliant and are automatically connected to the gateway with a discovery mechanism. When we installed the smart home in 2011,  the devices at that time provided little human-readable information, so it was very difficult to match the objects on the computer with the real devices in the home. When the connection to the devices was lost due to trouble, it was not possible to obtain information such as which device was the problem, where it was, who installed it, and when. After that, the solution is to additionally describe this information to the device descriptions managed by the gateway. These items listed in the core data model provide this information.

benfrancis:
@mryuichi wrote:
> I don't think it's necessary to delete all the items in the core data model session. The reason is as follows.
...

These are all valid points and I agree that there are many benefits to users of Thing Descriptions providing human readable strings which can be used in a user interface. However, I maintain that human readability is a separate goal to interoperability. As I said in https://github.com/w3c/wot-profile/pull/125#issuecomment-966589941 in the specification the goals are listed as:

1. guarantee interoperability among all implementations of the profile.
2. limit the implementation complexity for resource constrained devices.
3. ensure that a thing description is also useful for a human reader.

We agreed in #73 to drop goals 2 and 3, and focus only on goal 1.

If you think we should re-instate goal number 3 then I think we need a group decision to arrive at a consensus on that. See #138

mlagally:
@mryuichi I completely agree with your feedback.
Unfortunately we have seen many cases in the past decade, where devices became unmanageable due to the lack of human readable documentation.
Ultimately the consumer of a WebThing is going to be a human being, who is either using a thing via a UI (automatically generated from the TD), implementing a consumer for it (he also needs to understand what the affordances *mean* to use them.
@benfrancis 
We agreed that it is not the primary goal to enforce human readable TDs, but you put it almost as a non-goal, which will be difficult to reach consensus on.
Also the second point about reducing implementation complexity (for small embedded devices). We gave up on focusing on small embedded devices as consumers, but reducing / minimizing implementation complexity is still a goal of *all* profiles.

I think we should make the goals more precise:

- guarantee interoperability among all implementations of the profile.
- limit the implementation complexity.

let's fork out the discussion on usability for humans to  #138

benfrancis:
@mlagally wrote:
> let's fork out the discussion on usability for humans to #138

OK good idea, I have left my comments there.

@sebastiankb wrote:
> The removing of the whole chapter seems a bit harsh to me.

I was looking back through the commit history and it appears to me that the majority of this chapter was never actually formally reviewed before it was initially landed. If that turns out to be the case, would it change your view?

Please correct me if I'm wrong @mlagally, but most of this text appears to have been added in PR #1 as part of approx. 1,500 lines of text that don't seem to have been subject to the usual review process. Perhaps there was a resolution in a call to land all of those normative assertions without comment, but I can't find any reference to that. Apologies if I've misunderstood and this was subject to formal review elsewhere before the wot-profile repo was created.

If this text has not in fact yet been formally reviewed then I think that further supports the argument that all of the content of this chapter (and potentially others) should be reconsidered, not only in light of the revised goals and scope of the specification, but also so that it can be properly reviewed with the same level of scrutiny as other contributions.

With a view to finding a way forward, how about we start by commenting out the content of section 5.1, then create PRs to uncomment it sub-section by sub-section so that each part can be properly reviewed? I already provided section-by-section comments in https://github.com/w3c/wot-profile/pull/93#pullrequestreview-717982575 back in July, but given the lack of response perhaps that was too much feedback to be considered in a single PR and it needs to be broken down in order for us to make progress.

index.html
```diff
@@ -450,982 +450,33 @@ <h2>Profiling Mechanism</h2>
   <!-- Core Profile -->
   <section id="core-profile">
     <h2>WoT Core Profile</h2>
```
mlagally, line 452: This kind of huge PR cannot be discussed and reviewed in reasonable time.
The Data Model section will be significantly reworked in https://github.com/w3c/wot-profile/pull/93, so this MR should not be merged.

The proposed new text about discovery and link relation types should be carried over to individual MRs to enable timely reviews.

mlagally (changes requested):

index.html
benfrancis: @mlagally wrote:
> This kind of huge PR cannot be discussed and reviewed in reasonable time.

There are only 25 lines of new text in this PR, which just provides a list summarising the proposed updated contents of the Core Profile.

The rest is removing what appears to be mostly un-reviewed text which I think should be removed and then re-added piece by piece with the proper level of scrutiny. I note that you have not responded to my proposal in https://github.com/w3c/wot-profile/pull/125#issuecomment-977978088.

> The Data Model section will be significantly reworked in #93, so this MR should not be merged.

@egekorkan and I have already commented extensively on #93 to explain why it doesn't address our concerns. 

> The proposed new text about discovery and link relation types should be carried over to individual MRs to enable timely reviews.

The discovery section has already been split out into a separate PR (#121).

There is an ongoing discussion about a potential links section in #114.

There is just a single sentence about each of those topics in this PR, as part of the list of contents, which is intended to show which parts from the of the Core Profile I propose to keep. I am happy to remove those sentences if those sections don't end up being created, but that doesn't seem to be the biggest issue here.

mlagally:
> Please correct me if I'm wrong @mlagally, but most of this text appears to have been added in PR #1 as part of approx. 1,500 lines of text that don't seem to have been subject to the usual review process. Perhaps there was a resolution in a call to land all of those normative assertions without comment, but I can't find any reference to that. Apologies if I've misunderstood and this was subject to formal review elsewhere before the wot-profile repo was created.

In yesterday's call I explained the process how the initial draft has been created based on a strawman I provided as a starting point of the profile work. We had numerous conversations and did a formal WG review before we published the first WD and a resolution for publication. 

You joined the call half a year ago and there were many WG calls with a lot of ideas and discussions. 
If you are interested in the details, discussions  and work that happened in the 1.5 years before you joined, I recommend to review the call notes of these calls.

relu91 (commented):
After reviewing the content, comment section, and related issues (specifically https://github.com/w3c/wot-profile/issues/73 and https://github.com/w3c/wot-profile/pull/93) of this PR I would like to add a few words.

First of all, to an external reviewer, it is not clear IF the goals and objectives of the Core Profile document were really redefined. Even if @benfrancis mentioned in almost every comment that now the scope of the document is "just" interoperability, linking to https://github.com/w3c/wot-profile/issues/73 as the source, what I'm seeing there is:
1. The issue is still open
2. There's no single statement that says: as the group we agreed that the new scope is ... etc.
Furthermore, the discussion in this PR is going around about human readability in at least three comments: 
1. https://github.com/w3c/wot-profile/pull/125#issuecomment-966480260
2. https://github.com/w3c/wot-profile/pull/125#issuecomment-972841613
3. https://github.com/w3c/wot-profile/pull/125#issuecomment-974226503 (which branch the discussion in yet another dedicated issue: https://github.com/w3c/wot-profile/issues/138)

If this agreement has been tracked as a resolution in one of the calls (which probably is) I would strongly recommend closing the https://github.com/w3c/wot-profile/issues/73 with one final comment explaining the consensus. This would help others to understand the final goal of the Core Profile quicker and focus on the actual content of the PR. 

Putting that aside and taking the assumption that the goal has been refined, I would approve this PR. I'm a fan of the Single Responsibility Principle and I think too that the previous text was addressing multiple goals within the same document. I'm also advocating that text was mostly covering human readability and constraint devices concerns, 

The text is not really forever lost, on the contrary, I would advocate for keeping it in mind in other sections/documents within the group. For example, the human-readable information maybe even moved to a possible linting section in the Thing Description document (See https://github.com/w3c/wot-thing-description/issues/1195). I will expand this point in https://github.com/w3c/wot-profile/issues/138). 

### Conclusion
✔ I would approve this PR if the aforementioned goals are clearly shared between the group members.

benfrancis:
@relu91 wrote:
> it is not clear IF the goals and objectives of the Core Profile document were really redefined.

You're right, I agree it isn't completely clear from the comments what the conclusion was. I have proposed a resolution in https://github.com/w3c/wot-profile/issues/73#issuecomment-986815478 to formalise that decision.

@mlagally wrote:
> In yesterday's call I explained the process how the initial draft has been created based on a strawman I provided as a starting point of the profile work. We had numerous conversations and did a formal WG review before we published the first WD and a resolution for publication.
> 
> You joined the call half a year ago and there were many WG calls with a lot of ideas and discussions before. If you are interested in the details of the work that happened in the 1.5 years before you joined, I recommend to review the call notes of these calls.

Perhaps it is considered OK that the 1,500 lines of text in PR #1 did not have a detailed review or resolution before merging because it was only a "strawman proposal", but the idea of a [strawman proposal](https://en.wikipedia.org/wiki/Straw_man_proposal) is that it isn't the final text.

I didn't attend WoT Architecture calls in 2019 (I wasn't invited), but I and others have provided feedback throughout the whole process regarding the conflicting requirements and problems with the current text. That feedback dates back to the WoT Workshop in Munich in June 2019, where to my knowledge the idea of profiles was discussed for the first time, not just the last six months. Some notable examples:

- **2019-08-29** - @danielpeintner filed #3 expressing a concern that some of the kinds of proposed data schema constraints (e.g. only allowing one level of nesting) would be too restrictive
- **2019-08-30** @mmccool filed #7 , raising concerns regarding the conflicting requirements of the three proposed goals, suggesting that the WoT Profile specification focus only on interoperability and not human readability or resource constrained devices
- **2019-09-02** Minutes from the 2019-08-29 WoT Architecture call were published which shared the link to the strawman proposal in the W3C repo for the first time. The same day, I filed issue #10 suggesting that "section 4.1.1 Core Data Model should be removed, or merged with the main WoT Thing Description specification". The issue didn't receive a response.
- **2020-12-18** - Following the publication of the First Public Working Draft I provided detailed feedback in an [email](https://lists.w3.org/Archives/Public/public-wot-wg/2020Dec/0025.html) to the mailing list. I filed #59, #60 and commented on issue #10, again highlighting the conflicting requirements and saying that if the WoT Core Data Model section was not removed then the number of constraints should at least be drastically reduced.
- **2021-03-18** @mmccool raised issue #73, again raising the issue of conflicting requirements and suggesting refining the goals and scope of the specification. After a long discussion Sebastian asked that I create a PR with the new document structure I had proposed.
- **2021-04-22** - I created #78 for the part of the restructure which involved moving parts out of the WoT Core Data Model section and removing the remaining parts
- **2021-07-29** - I provided [detailed section by section feedback](https://github.com/w3c/wot-profile/pull/93#pullrequestreview-717982575) on #93 explaining why my recommendation is to "remove the entire section and redistribute parts of its content to other sections of the WoT Profile specification"
- **2021-11-04** - @mlagally asked me to split #78 into smaller PRs so I split it into 6 smaller PRs, including this one.

Given @relu91's comment I think the first step is probably to make absolutely sure we have agreement on the goals and scope of the specification. Once we are clear on the scope, I suggest that the WoT Core Data Model section (and other sections) need to be re-evaluated to ensure the text still fits within that scope. My view remains that the issues with this text (the majority of which dates back to the original strawman proposal) are significant enough that the best approach would be to remove or comment out the entire section and then re-review section by section to see which areas we want to keep.

mlagally:
Arch call on 11.5.:
This PR has merge conflicts and won't be easy to fix, so we agreed to close it. 
Let's discuss the topic it tries to resolve separately and create a new PR, if required.

Issue #120, opened 2021-11-08
Complete or remove JSON Schema of the Core Profile


benfrancis:
There's currently an incomplete section of the specification called [JSON Schema of the Core Profile](https://w3c.github.io/wot-profile/#core-profile-json-schema).

This section either needs to be completed, or removed, before publication.

Note that if the WoT Core Data Model section is reduced or removed as I have proposed in #10 then this section may be of limited value and the JSON schema provided by the WoT Thing Description specification would be sufficient.

egekorkan:
I think there will be and has to more restrictions that the schema of the TD. For example, checking for `http://` in form `href` values.

mlagally:
I completely agree with @egekorkan that we need a JSON Schema for automatic validation.
The primary goal of the Profile spec is to drive interoperability and we need additional constraints that make that happen.
The added value of the Profile is way more than just providing an unambiguous the HTTP protocol binding.

mlagally:
tentatively assigning to @egekorkan and @FadySalama and hoping to get their help.

egekorkan:
I agree that the contraints should be agreed on first since there is discussion on nearly every detail. After that, we can write a script that takes the TD schema creates another schema that has more constraints

mlagally:
Defer to Profile 1.1.

benfrancis:
Note that the Core Profile now no longer exists. JSON schemas could be created for each published profile, but that may have limited value because profiles can not be completely described by a JSON schema.

egekorkan:
I would still see the value of having a JSON Schema for each profile, especially considering that they will be a subset of TD later on. However, a set of linting rules like with [Spectral](https://stoplight.io/open-source/spectral) might fit better since a JSON Schema will always do a strict validation (which is too assertive in some cases)

PR #133, opened 2021-11-16
Reinstating the Open Issues section

mlagally:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/133.html" title="Last updated on Nov 16, 2021, 9:39 AM UTC (e5876b3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/133/316ba88...e5876b3.html" title="Last updated on Nov 16, 2021, 9:39 AM UTC (e5876b3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2619,6 +2619,14 @@
     <p class="ednote">
       Todo: Define a JSON-SCHEMA.
     </p>
+</section>
+<section id="open-issues">
+    <h3>Open Issues</h3>
+    <ul>
+      <li>define a JSON schema</li>
+      <li>Examples</li>
+    </ul>
+  </section>
 
 </body>
 


benfrancis:
Hi @mlagally, the reason I removed this section is because it was out of sync with the issue tracker on GitHub, which is likely to keep happening.

If you'd like to let readers know about open issues, perhaps this section could just include a link to the [issue tracker](https://github.com/w3c/wot-profile/issues), or perhaps the [blocks-publication](https://github.com/w3c/wot-profile/labels/blocks%20publication) label that you created? That includes the [JSON schema issue](https://github.com/w3c/wot-profile/issues/120) for example.

Note that there is already a link to the issue tracker in the Repository section of the `otherLinks` metadata of the specification, which gets rendered at the top of the page (The Thing Description specification has an "open issues" link in a Feedback instead).

mlagally:
@benfrancis
I like the idea of having an additional reference to the blocks-publication labels in github, however I think a section in the document has additional value.
We should review the issues that are marked with the label and agree as a group that they block publication. 
We may have issues labeled as blockers which are not critical for publication and need to get group consensus.

Issue #137, opened 2021-11-18
Geolocation


mlagally:
The profile should define standard metadata properties for location.
This is required for interoperability with sensors.

benfrancis:
See also: https://w3c.github.io/wot-thing-description/#semantic-annotations-example-geoloc

mlagally:
The current example in the TD does not contain an altitude, should be added there.
The profile could RECOMMEND to use the vocabulary from the TD example:  (https://www.w3.org/2003/01/geo/)

benfrancis:
I agree that recommending a standard way to express geolocation in the Core Profile would aid interoperability.

I think the consensus around the Thing Description specification is that existing external ontologies like schema.org should be used for these kinds of metadata, rather than defining something new in the Thing Description namespace. I agree with this.

Where I think this gets a bit tricky is if we're going to recommend a particular external ontology for semantic annotations for geolocation, should we be recommending a particular ontology for other concepts too? We've already discussed units as another example (#29). Where do we stop?

Schema.org defines ontologies for lots of other concepts that could be relevant to IoT devices, and iotschema.org (whilst not as stable) goes even further. There are also lots of other schema repositories like oneiota.org, onedm.org and webthings.io/schemas.

What consistent criteria could we define to decide what should and shouldn't be recommended as part of the Core Profile?

mlagally:
Deferring to Profile 1.1, since we currently have different opinions on where/how geolocation should be handled.

benfrancis:
Please note closed pull request https://github.com/w3c/wot-profile/pull/264 for when we re-visit this in 1.1.

PR #149, opened 2021-11-30
Remove vocabulary note - closes #17

### LINKED ISSUE #17: Vocabulary not in TD context
[Section 4.1.2.2](https://w3c.github.io/wot-profile/#recommended-practice- ) list a number of vocabulary (e.g., softwareRevision, loc_latitude,...) that are not mention in the TD specification yet. 

Is there an intention to define a new context file that defines these terms or do you refer to existing ontologies that use these terms?

benfrancis:
***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/149.html" title="Last updated on Nov 30, 2021, 4:10 PM UTC (21fcf86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/149/316ba88...benfrancis:21fcf86.html" title="Last updated on Nov 30, 2021, 4:10 PM UTC (21fcf86)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -683,22 +683,6 @@
             address SHOULD be used.
             </span>
           </p>
-
-          <section class="note">
-            <p>
-              <span class="rfc2119-assertion" id="profile-5.1.2.1-thing-mandatory-fields-4">
-              It will be evaluated whether the profile also recommends some new TD terms that may be
-              introduced in TD 1.1.
-              </span>
-              Currently the following terms are discussed: serialNumber,
-              hardwareRevision, softwareRevision, loc_latitude, loc_longitude
-              loc_altitude, loc_height, and loc_depth.
-              <span class="rfc2119-assertion" id="profile-5.1.2.1-mandatory-fields-5">
-              If these, or some of them, are defined in the TD
-              1.1 model, they may be recommended here in one of the next draft updates.
-              </span>
-            </p>
-          </section>
         </section>
       </section>
 


mlagally:
The note is outdated and we should have replacement language that gives guidance of how to encode geolocation and hardware details (version, serial, ...)

mlagally (commented):
This text removes terms that were to be included into the TD spec.

This was a precondition for removing the previous section from the profile which happened about a year ago.

I just checked the geolocation case, which is very important for interoperability:
Unfortunately the TD specification does not define corresponding vocabulary terms for these values and just contains an example that references a rather old IG note.

Therefore I suggest to hold off deleting this section until we have a clear way to specify these attributes in the profile.

mlagally (changes requested):
The current text was a placeholder for specification language we expected from the TD specification. 
It should be kept as a placeholder until we have appropriate replacement language, that enables to define "profile-things" that have a location and common metadata that allow 
common management of these (e.g. displaying them on a map, handling hardware and software version information).

These additional fields are one of the key benefits of the profile, since they enable interesting interop scenarios, that are not possible without them.

I strongly suggest to decide on the key names and agree on a format for the values.
If we cannot agree on a single format for profile purposes, we could define an unambiguous way of indicating the interpretation of the values.

benfrancis:
My understanding of the action from the architecture call [noted on 25th November 2021](https://github.com/w3c/wot-profile/issues/17#issuecomment-979144471) was that the decision was to just remove the note.

I don't agree with the need to extend the Thing Description specification with `serialNumber`, `hardwareRevision`, `softwareRevision`, `loc_latitude`, `loc_longitude` `loc_altitude`, `loc_height`, and `loc_depth` members inside the WoT Profile specification*. I've therefore unassigned myself from issue #17 so that if someone wants to make the case for those members they can assign themselves to the issue.

I'll close this PR since it doesn't seem to be needed.

\* For discussion about an alternative approach for geolocation data, see #137.

Issue #134, opened 2021-11-17
Structured event object


mlagally:
As discussed in issue https://github.com/w3c/wot-profile/issues/126 we need to define a structured event object and not use protocol specific workarounds.

benfrancis:
In my opinion this would be an unnecessary over-complication. See https://github.com/w3c/wot-profile/issues/126#issuecomment-971507879 and the previous discussion in #100.

mlagally:
@benfrancis 
After rereading the discussions in #126 and #100, I still believe we have a significant disconnect here.
We are defining a common HTTP protocol binding with JSON payloads.
 At this moment we have a single subprotocol for events, which does not have a standardized way to include event metadata. 

The SSE based event model has additional limitations - It does work well for some use cases, specifically for home gateway scenarios or scenarios in a local network.
However the current event model is not scalable to a variety of use cases of a HTTP profile.
We need other event mechanisms that are able scalable and are able to pass through firewalls.

I'm convinced that instead of a sub-protocol specific encoding of metadata in the id, a common structured JSON based event format across different event subprotocols is required.

Please don't think of browsers as the only consumers - in industrial scenarios the communication partner is a server, for which an endpoint for WebHooks is a natural way to expose a scalable messaging interface.

benfrancis:
@mlagally wrote:
> We are defining a common HTTP protocol binding with JSON payloads.

Agreed.

> At this moment we have a single subprotocol for events, which does not have a standardized way to include event metadata.

The specification _does_ [define a standardised way to include event metadata](https://w3c.github.io/wot-profile/#core-profile-protocol-binding-events-subscribeevent-6):

>> Whenever an event of the specified type occurs while the Web Thing has an open connection with a Consumer, the Web Thing MUST send event data to the Consumer using the event stream format in the Server-Sent Events [EVENTSOURCE] specification. For each message sent, the Web Thing MUST set the event field to the name of the EventAffordance and populate the data field with event data, if any. The event data MUST follow the data schema specified in the EventAffordance and MUST be serialized in JSON. The id field SHOULD be set to a unique identifier for the event, for use when re-establishing a dropped connection (see below).

Event data is serialised in JSON based on the data schema from the event affordance and provided in the `data` field of the event. The event name is provided in the `event` field and a timestamp can be provided in the `id` field (which is common practice in SSE).

This is not a "workaround", it is a clean mapping of the WoT Information Model onto the Server-Sent Events protocol.

What do you think is missing?

[...]

Issue #126, opened 2021-11-10
Dependency on Cloud Events


egekorkan:
@mlagally has mentioned the F2F before TPAC of using cloud events in the profile and there are also other mentions of it in various issues. Something that needs to be documented somewhere is that the CloudEvents specification is driven by the Cloud Native Foundation, which is not a standards organization, thus cannot be simply linked to. This implies that having a dependency requires this profile spec to mirror their specification, as TD does with JSON Schema. The group should decide on whether this work is feasible in the current charter before deciding on adopting this specification.

mlagally:
@egekorkan 
Thanks for flagging the problem. 
I think we do not have to reference the entire cloud event specification (which is not large btw.) but just select the parts that are relevant for us. As explained in #100, this is only a very small section, which could be just included in the same way.

benfrancis:
Apart from not being from a standards body (whether in part or in full), I don't think a dependency on CloudEvents is needed or desirable for the Core Profile, see #100.

mlagally:
@benfrancis 
I agree, we don't need the full cloud events, however a structured JSON payload with clear properties to be useful. Tying these into specifics of the the protocol binding is not desirable.
I think the event object should be protocol agnostic. 

Wrt. reference problem: we can resolve by defining a compatible payload as part of the profile.

benfrancis:
> Tying these into specifics of the the protocol binding is not desirable.
>  think the event object should be protocol agnostic.

Again, "tying into specifics of the protocol" is the whole point of a protocol binding. Protocol bindings should not be protocol agnostic because their purpose is to be protocol specific. Other event mechanisms in other protocols may also have their own payload formats and may use a non-JSON serialisation or binary data format, so there's not even any guarantee that the same JSON wrapper could be re-used.

The current event protocol binding provides a very simple and neat mapping from the WoT information model onto the server-sent events event stream format:

1. The event name maps onto the `event` field
2. The data schema maps onto the `data` field
3. A timestamp can optionally be provided in the `id` field

Example Thing Description:
```
  "events": {
    "overheated": {
      "data": {
        "type": "number"
      },
      "forms": [{
        "href": "./events/overheated",
        "subprotocol": "sse"
      }]
    }
  },
```

Example event payload:
```
event: overheated\n
data: 90\n
id: 2021-11-16T16:53:50.817Z\n\n
```

[...]

Issue #107, opened 2021-10-19
Events - Scalable event mechanism for cloud use cases


mlagally:
The current event mechanism in the core profile relies on SSE.

Whereas this is completely appropriate for deployments, where only a few devices are managed by a clonsumer, this mechanism does not scale well to scenarios, where hundreds or thousands of devices are sending events to a common consumer.

The SSE mechanism is basically a "poll" with an open connection. For devices with a low rate of event notifications, this is not appropriate, since it consumes network resources without significant traffic.

We need a second "push" based mechanism that can be used in deployment scenarios of hundreds or thousands of devices.
Webhooks are a natural choice for device initiated event notifications.
The TD specification already contains a Webhook example: https://w3c.github.io/wot-thing-description/#webhook-example-serialization

The profile specification has to define a corresponding section.

benfrancis:
> this mechanism does not scale well to scenarios, where hundreds or thousands of devices are sending events to a common consumer.

There are certainly more efficient ways of pushing events over a network (like UDP multicasting), but not necessarily over HTTP (the protocol used by the Core Profile).

> The SSE mechanism is basically a "poll" with an open connection.

That's not how I'd describe it. it's a push mechanism over a TCP socket. It's a special type of HTTP request which the server knows to keep open, as opposed to long-poll where it has to be re-opened every time a message is pushed to the client. In this respect it's similar to WebSockets except that it's unidirectional, limited to UTF-8 encoded messages, has some limited built-in semantics for describing events, and its own dedicated JavaScript API in web browsers.

> We need a second "push" based mechanism that can be used in deployment scenarios of hundreds or thousands of devices.
> Webhooks are a natural choice for device initiated event notifications.

I disagree.

Firstly, do you have evidence that webhooks would be any more efficient? Have you tested it? Webhooks would require keeping fewer TCP sockets open, but those may be relatively cheap compared with needing to send a whole new HTTP request for every event emitted.

Secondly, webhooks require that the consumer is a web server which is not always (I'd say usually not) the case. What if the consumer is a local desktop or mobile app for example?

> The profile specification has to define a corresponding section.

Is that a resolution, or your opinion?

egekorkan:
+1 to all the points above by @benfrancis

mlagally:
The fundamental problem in these "poll"-based event mechanism - or should we say "consumer initiated" mechanisms is that they don't match to scenarios behind a firewall. 

Wrt. server deployment scenarios:
[...]

PR #150, opened 2021-12-01
event-payload-format

mlagally:
Structured event format that ensures interoperability of event payloads with the cloud events specification. (https://github.com/cloudevents/spec/blob/v1.0.1/spec.md)

Additional event bindings, such as for observable properties and bindings to other HTTP/JSON based event subprotocols should use the same format to ensure interoperability.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/150.html" title="Last updated on Dec 1, 2021, 7:30 AM UTC (b6cf9e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/150/316ba88...b6cf9e1.html" title="Last updated on Dec 1, 2021, 7:30 AM UTC (b6cf9e1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -94,9 +94,9 @@
   </script>
   <style>
     .rfc2119-assertion {
-      background-color: rgb(230,230,230)
+      background-color: rgb(230, 230, 230)
     }
-    </style>
+  </style>
 
 </head>
 
@@ -112,9 +112,9 @@
       to use the device in a certain scenario.
       These actions can be done by anyone without specific training.
     </p>
-      The <a>WoT Core Profile</a> defines a set of <em>constraints
+    The <a>WoT Core Profile</a> defines a set of <em>constraints
       and rules</em>, which compliant thing descriptions have to adopt
-      to guarantee interoperability.
+    to guarantee interoperability.
     </p>
     <p>
       These rules are prescriptive, to ensure that
@@ -147,8 +147,8 @@
         </p>
         <p>
           <span class="rfc2119-assertion" id="profile-abstract-1">
-          A TD that is compliant to the <a>core profile</a> MUST adhere to
-          both the constraints on the data model and the protocol binding.
+            A TD that is compliant to the <a>core profile</a> MUST adhere to
+            both the constraints on the data model and the protocol binding.
           </span>
         </p>
       </li>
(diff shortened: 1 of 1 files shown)


benfrancis (changes requested):
Hopefully it should come as no surprise to you that I don't approve of this change, for all the reasons already discussed in #100, #107, #126 and #134.

I am baffled as to why you continue to pursue this when it's clear that the consensus of the group so far is that it's not a good idea. There is a far simpler solution to the timestamp requirement proposed in #119 which only requires a one line change to the specification and much simpler data schemas and event payloads.

I have already implemented the current event protocol binding defined in the specification in WebThings Gateway and another Krellian product and I have no intention of changing them to adopt this unnecessarily complicated and over-engineered solution. I'm afraid you will need to find two other implementations willing to adopt this.

egekorkan:
If we want to align with cloud events, we should simply have a binding template for it and ideally some keyword like `subprotocol` to indicate the alignment in TDs. A profile can prescribe that this subprotocol must be used. The metadata around the protocol should not be put into the TDs, it would be bad practice from our side. 

To better illustrate my point with an analogy, it would be saying that the payload in a profile X should contain the data schema of the corresponding affordance.

mlagally:
@egekorkan 
Defining a subprotocol and a binding template for an event payload looks to me as a significant overhead to achieve only a simple goal, i.e. having a structured payload that some other cloud systems are already using.

I think it is appropriate to define the payload scheme as part of the event binding.

egekorkan:
> Defining a subprotocol and a binding template for an event payload looks to me as a significant overhead to achieve only a simple goal, i.e. having a structured payload that some other cloud systems are already using.

> I think it is appropriate to define the payload scheme as part of the event binding.

I agree with this. However, I strongly object to proposing cloud events for a WoT green field scenario. There should be a binding for cloud events anyway given the fact that you are saying it is heavily used. If we are the WoT standardization group and are proposing a new binding (this is a binding by the way since you are using an existing "standard") for people to implement in a given profile that is about WoT, we should leverage the existing technologies we have in WoT, namely the TD. 

If you write an example TD for this, I can better illustrate why this is a bad design choice.

PR #157, opened 2022-01-10
Define protocol binding for observing properties - closes #95

### LINKED ISSUE #95: Should the Core Profile define a protocol binding for `observeproperty`/`unobserveproperty`?
Question raised by @egekorkan in https://github.com/w3c/wot-profile/pull/92#issuecomment-890259947:
> * I think there was a decision to not support `observeproperty` (and its similar friends). Is this decision still valid if this PR gets accepted?

I personally don't mind either way. Subscribing is the primary (default) use case of an event affordance, whereas observing is arguably more of a nice-to-have for a property. I wouldn't be opposed to defining a protocol binding for `observeproperty`/`unobserveproperty` using Server-Sent Events (to reduce implementation complexity) as an optional feature of the Core Profile.

I think it could be fairly straightforward using HTTP content negotiation to choose between reading a static JSON value vs. observing an event stream, we'd just need to decide what to use as the event name (use the property name?).

benfrancis:
This is a proposed protocol binding for observing properties in the Core Profile, as discussed in #95. It includes bindings for the following WoT operations:
- `observeproperty`
- `unobserveproperty`
- `observeallproperties`
- `unobserveallproperties`

(There are also a few minor fixes to the HTML markup for the events protocol binding  which I noticed along the way).

This protocol binding uses the Server-Sent Events mechanism to push property changes to a Consumer, consistent with the events protocol binding.

Using the Server-Sent Events mechanism for observing properties isn't as obvious as using it for subscribing to events, but if a Consumer is already implementing SSE for events then it seems silly to use a completely different push mechanism for observing properties.

An alternative to Server-Sent Events would be long-polling, but this has many drawbacks:
- No built-in mechanism for re-establishing dropped connections
- No built-in mechanism for catching up on missed data
- No native API or network optimisations built into web browsers
- Higher load on the Producer/server due to repeated HTTP requests

Without this added feature currently the only way to observe property changes within the Core Profile is to manually poll a property URL with HTTP GET requests, which is very inefficient.

I would value feedback on this proposal.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/157.html" title="Last updated on Jan 10, 2022, 5:23 PM UTC (36b2338)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/157/707eee7...benfrancis:36b2338.html" title="Last updated on Jan 10, 2022, 5:23 PM UTC (36b2338)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1583,6 +1583,159 @@
           HTTP/1.1 204 No Content
           </pre>
         </section>
+
+        <section id="observeproperty">
+          <h5><code>observeproperty</code></h5>
+          <p class="rfc2119-assertion" 
+            id="core-profile-protocol-binding-properties-observeproperty-1">
+            The URL of a <code>Property</code> resource to be used when
+            observing the value of a property MUST be obtained from a Thing 
+            Description by locating a
+            <a href="https://www.w3.org/TR/wot-thing-description/#form">
+            <code>Form</code></a> inside the corresponding
+            <a href="https://www.w3.org/TR/wot-thing-description/#propertyaffordance">
+            <code>PropertyAffordance</code></a>
+            for which:
+          </p>
+          <ul class="rfc2119-assertion" 
+            id="core-profile-protocol-binding-properties-observeproperty-2">
+            <li>
+              Its <code>op</code> member contains the value 
+              <code>observeproperty</code>
+            </li>
+            <li>
+              After being resolved against a base URL where applicable, the
+              <a href="https://tools.ietf.org/html/rfc3986#section-3.1">URI
+              scheme</a> [[RFC3986]] of the value of its <code>href</code>
+              member is <code>http</code> or <code>https</code>
+            </li>
+            <li>Its <code>subprotocol</code> member has a value of
+              <code>sse</code>
+            </li>
+          </ul>
+          <p class="rfc2119-assertion" 
+          id="core-profile-protocol-binding-properties-observeproperty-3">
+            The resolved value of the <code>href</code> member MUST then be used
+            as the URL of the <code>Property</code> resource.
+          </p>
+          <p class="rfc2119-assertion" 
+            id="core-profile-protocol-binding-properties-observeproperty-4">
+            In order to observe a property, a Consumer MUST follow the 
+            Server-Sent Events [[EVENTSOURCE]] specification to open a
+            connection with the Web Thing at the URL of the 
+            <code>Property</code> resource.
+          </p>
+          <p>
+            This involves the Consumer sending an HTTP request to
+            the Web Thing with:
+          </p>
+          <ul>
+            <li>Method set to <code>GET</code></li>
+            <li>URL set to the URL of the <code>Property</code> resource</li>
+            <li>
+              <code>Accept</code> header set to <code>text/event-stream</code>
+            </li>
+            <li>
+              <code>Connection</code> header set to <code>keep-alive</code>
+            </li>
+          </ul>
+          <pre class="example">
+            GET /things/lamp/properties/level HTTP/1.1
+            Host: mythingserver.com
+            Accept: text/event-stream
+            Connection: keep-alive
+          </pre>
+          <p class="note" title="Opening a connection using JavaScript">
+            For Consumers implemented in JavaScript [[ECMASCRIPT]] and executed 
+            in a runtime which exposes the
+            <a href="https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface">
+            <code>EventSource</code></a> interface, a Server-Sent Events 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1583,6 +1583,159 @@ <h5><code>writeproperty</code></h5>
...
+            and following the data schema specified in the 
+            <code>PropertyAffordance</code>. The <code>id</code> field SHOULD be set to a unique 
+            identifier for the property change, for use when re-establishing a
+            dropped connection (see below). It is RECOMMENDED that the identifier
+            is a timestamp representing the time at which the property changed, in
+            the &quotdate-time&quot format specified by [[RFC3339]].
+          </p>
+          <pre class="example">
+            event: level\n
+            data: 42\n
```
sebastiankb, line 1701: I think two "\n" characters are needed to end the stream.

sebastiankb: Maybe a JSON-based serialized data would be suitable to show here, or?

benfrancis: Thanks for the review.

@sebastiankb wrote:
> I think two "\n" characters are needed to end the stream.

As I understand it the double carriage return comes after the last field in the message to signify the end of the message, so it's only needed after the `id` field in this case.

> Maybe a JSON-based serialized data would be suitable to show here, or?

Technically this _is_ serialised JSON :) That might be more obvious with a property with an object type, but both the properties in the example Thing used throughout the protocol binding section are simple types (a boolean and an integer). Changing that would require changing all the examples. Alternatively we could add additional more complex examples in an appendix?

mlagally: > > Maybe a JSON-based serialized data would be suitable to show here, or?
> 
> Technically `42` _is_ serialised JSON :) That might be more obvious with a property with an object type, but both the properties in the example Thing used throughout the protocol binding section are simple types (a boolean and an integer) - see the Thing Description in #91. Changing that would require changing all the examples. Alternatively we could add additional more complex examples in an appendix?

I would encourage to think of complex properties when we design this protocol binding. Of course a single value property is trivial, and the event payload structure as well. If we delegate considerations of complex properties and cases to examples, we open the door widely for different interpretations.
For a solid specification we have to precisely define the behaviour of consumers for these cases, and I strongly support a structured event format that resolves these issues in a clear and simple way.

Please review https://github.com/w3c/wot-profile/pull/150 (again), which provides this kind of format. It is self contained specification text, i.e. does not have any dependency from other specifications, however is aligned with the CloudEvents spec to provide interop with that widely used spec.

mlagally (changes requested):
The SSE event mechanism, with the underlying assumption that all connections are initiated by the consumer, blocks consumer resources (network connections) and causes protocol overhead (keep alive messages).
It is limited to use cases where a consumer (e.g. browser client) is interactacting with a few things.

It does not support battery powered devices (draining the battery due to keep alive network traffic) and does not scale to deployments with thousands of devices.

It is therefore not sufficient for critical use cases.
To enable these, we need another event mechanism, where the network connection is established by the thing only in cases, when there's an actual event.

The event payload format should be a structured JSON payload, as proposed in https://github.com/w3c/wot-profile/pull/150

index.html
benfrancis: > I would encourage to think of complex properties when we design this protocol binding. Of course a single value property is trivial, and the event payload structure as well. If we delegate considerations of complex properties and cases to examples, we open the door widely for different interpretations.

There is no room for interpretation in the way that JSON is serialised and de-serialised from a string, that is defined in [RFC8259](https://www.rfc-editor.org/rfc/rfc8259).

An integer would be serialised in an event payload as follows:
`data: 42\n`

A more complex object would be serialised as follows:
`data: {"question":"The meaning of life, the universe and everything","answer":42}\n`

The purpose of a more complex example would simply be to more clearly illustrate that it is in fact JSON (which is already specified in an assertion), not to show how it should be serialised.

> For a solid specification we have to precisely define the behaviour of consumers for these cases, and I strongly support a structured event format that resolves these issues in a clear and simple way.

What issues need resolving exactly?

> Please review #150 (again), which provides this kind of format. It is self contained specification text, i.e. does not have any dependency from other specifications, however is aligned with the CloudEvents spec to provide interop with that widely used spec.

As already discussed in relation to event affordances (in #100, #107, #126, #134 and #150), such a data wrapper would be redundant. As far as I can tell it adds no value, all it does increase the size of the payload. What specifically do you think is missing in the payload of a property change message? What is needed except the property name, the property value and an optional timestamp?

sebastiankb: @benfrancis 
>Technically 42 is serialised JSON :)

That's right :-)

mlagally:
Profile call on April 6th:
Needs to be retargeted to SSE section.

benfrancis:
Replaced by #191.

PR #181, opened 2022-03-02
Implementation Report and Assertion Markup Fixes

mmccool:
Draft implementation report.
Currently just a copy of the (also incomplete and draft) Architecture report.
Needs a lot of work.  There seems to be a bug also where only one assertion is getting picked up.

Updates: 
* the "bug" was due to some unexpected input, namely the rfc2119-assertion class being using with tags other than span
* Now getting all the assertions but they don't necessarily make sense; some have too many keywords, some have none, table assertions are not marked up...

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-profile/pull/181.html" title="Last updated on May 11, 2022, 11:01 AM UTC (ec18a38)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/181/cf7f5a9...mmccool:ec18a38.html" title="Last updated on May 11, 2022, 11:01 AM UTC (ec18a38)">Diff</a>


mmccool:
Good news:
- Implementation report now lists all the assertions in the document
Bad news:
- The assertion markup and wording had many issues that to fix will require some major changes to index.html
- This may cause various conflicts with other PRs in flight
- Issues included
    - using rfc2119-assertion in p and ul tags (only span is supported by my tool)
    - referring to external content with phrasing like "the following:" (if it is a table, then table rows should be marked as table assertions, and children of a parent assertion)
    - breaking up things into two different assertions, like a paragraph followed by a list, that are in fact one assertions (list items tending not to have RFC keywords)
    - having multiple RFC keywords and sentences in single assertion spans
    - using "NOT PERMITTED" which is afaik not an RFC2119 keyword

mmccool:
Anyhow, I've made a bunch of changes to the index.html but need to make more to fully resolve these issues.  In order for the tooling and implementation report generation to give reasonable results we need to have simple, testable, self-contained assertions with precisely one RFC2119 keyword, and need to mark rows in normative tables as assertions (see how the TD spec does this...)

benfrancis:
@mmccool Some of that is my fault, sorry. I intentionally removed redundant `<span>` tags because it's generally considered bad practice to litter HTML with semantically meaningless nested `<div>` and `<span>` tags.

I can look at fixing issues like multiple assertions in one tag or assertions split across multiple tags. I didn't realise you planned to automatically generate reports in this way when I wrote a lot of the markup, I thought the CSS classes were just used for rendering normative text with a different style.

Perhaps you could also improve your tool to support assertion classes inside more semantically meaningful tags like `<p>` where there's otherwise no reason to add a redundant `<span>`?

mmccool:
Well, I *could* improve my code, true, to pick up UL and P, but then I have to include the outer tag in my output, etc. which changes the code structure, will require debugging, etc.  So I'd rather not.  It's also easier for tools that don't really understand HTML that well to pick up spans, e.g. using grep.  I'm using cheerio so do not have that problem, but.

Anyhow, I've already gone through and added in a bunch of spans. I have a bit more cleanup to do but at this point would like to just press forward so we can get a CSV template we can use during the testfest.  The only problem is the breakage to other pending PRS.  I can try to put the spans on other lines to hopefully get the diffs to work out...

benfrancis:
> The only problem is the breakage to other pending PRS.

Yes unfortunately that "cleanup" is going to cause a lot of merge conflicts for the remaining 15 open pull requests, but nothing we can't resolve.

benfrancis:
Regarding creating an implementation report, are there actually any other implementations of the WoT Profile specification that the Working Group is aware of other than the [work-in-progress implementation](https://github.com/WebThingsIO/gateway/issues?q=is%3Aissue+is%3Aopen+label%3Awot-profile) in WebThings Gateway? If not then an implementation report may be a bit premature (though no harm in laying the groundwork).

danielpeintner:
> Regarding creating an implementation report, are there actually any other implementations of the WoT Profile specification that the Working Group is aware of other than the [work-in-progress implementation](https://github.com/WebThingsIO/gateway/issues?q=is%3Aissue+is%3Aopen+label%3Awot-profile) in WebThings Gateway?

[Eclipse thingweb node-wot](https://github.com/eclipse/thingweb.node-wot/) is compliant w.r.t. properties. It definitely behaves differently when dealing with actions & events. We, the thingweb team, needs to look into what it would require to be aligned with the profile.

mmccool:
OK, I rebased and fixed a lot of the issues, including assertions that included lists, and I also broke up a number of assertions that had multiple keywords.  There were indeed a lot of merge conflicts though, most caused by the change of a lot of the ids to use the new "http-" prefix rather than "core-".

Anyway... I would like to merge this ASAP so others can do the rebasing work, before it gets broken again.  As there won't be a profile call for a couple of weeks, I would like to call the question on this and do the merge after reviews via github, so that others will have time to fix their own merge conflicts before we meet again.

I also updated the implementation report, which was a useful sanity check (as I could see the list of assertions and check that they were broken up right).

Note some of the assertions went away so there are gaps in the ids, i.e. x-5 may be followed by x-7 if x-6 was in fact a list that was merged into x-5.  There were quite a few of these.

mmccool:
Please do check this carefully though.  I searched for "core-" and it turns out there are still a bunch that need cleaning up, but not any of the ones that I dealt with while resolving merge conflicts.

index.html
```diff
@@ -1833,9 +1867,6 @@ <h5><code>readallproperties</code></h5>
             <a href="https://www.w3.org/TR/wot-thing-description/#form-serialization-json">
             <code>forms</code></a> member
             for which:
-          </p>
-          <ul class="rfc2119-assertion" 
-            id="http-baseline-profile-protocol-binding-readallproperties-2">
```
benfrancis, line 1838: The `<ul>` tag is now missing

index.html
```diff
@@ -1987,7 +2024,8 @@ <h5><code>writemultipleproperties</code></h5>
         <h4>Actions</h4>
         <section id="http-baseline-profile-protocol-binding-invokeaction">
           <h5><code>invokeaction</code></h5>
-          <p class="rfc2119-assertion"
+	  <span class="rfc2119-assertion"
+	  <p>
```
benfrancis, line 2028: The `id` attribute needs to be inside the `<p>` tag.

index.html
```diff
@@ -2162,20 +2207,21 @@ <h6 id="ActionStatus"><code>ActionStatus</code> object</h6>
...
           <h6 id="sync-action-response">Synchronous Action Response</h6>
-          <p class="rfc2119-assertion" 
+	  <span class="rfc2119-assertion" 
             id="http-baseline-profile-protocol-binding-invokeaction-13">
+		  <p>
             If providing a Synchronous Action Response, a Web Thing MUST send 
             an HTTP response with:
-          </p>
-          <ul class="rfc2119-assertion" 
-            id="http-baseline-profile-protocol-binding-invokeaction-14">
```
benfrancis, line 2171: `<ul>` tag is now missing.

index.html
```diff
@@ -2263,43 +2313,51 @@ <h5><code>queryaction</code></h5>
...
+            id="http-baseline-profile-protocol-binding-queryaction-7a">
+            If the queried action failed to execute, then  
+            the <code>status</code> member of the <code>ActionStatus</code> object 
+            MUST be set to <code>"failed"</code>. 
+	  </span>
+	  <span class="rfc2119-assertion"
+            id="http-baseline-profile-protocol-binding-queryaction-7b">
+            If the queried action failed to execute, then  
+	    the <code>error</code>
+            member MAY provide additional error information
```
benfrancis, line 2358: I think this harms readability, but I guess it's necessary for automation.

index.html
```diff
@@ -2643,18 +2727,18 @@ <h3>Events</h3>
         </p>
         <section id="http-sse-profile-protocol-binding-subscribeevent">
           <h5><code>subscribeevent</code></h5>
-          <p class="rfc2119-assertion" 
+	  <span class="rfc2119-assertion" 
             id="http-baseline-profile-protocol-binding-subscribeevent-1">
```
benfrancis, line 2731: This id should be be `http-sse-profile-protocol-binding-subscribeevent-1`. That's not the fault of this PR, but probably easier to fix in this PR rather than create another merge conflict.

index.html
```diff
@@ -2736,64 +2821,80 @@ <h5><code>subscribeevent</code></h5>
...
+            id="http-sse-profile-protocol-binding-subscribeevent-7a">
             If the connection between the Consumer and Web Thing drops
             (except as a result of the <code>unsubscribe</code> operation 
             defined below), the Consumer MUST re-establish the connection 
             following the steps outlined in the Server-Sent Events specification
-            [[EVENTSOURCE]]. Once the connection is re-established the Web Thing
+            [[EVENTSOURCE]]. 
+	  </span>
+	  <span class="rfc2119-assertion" 
+            id="http-sse-profile-protocol-binding-subscribeevent-7a">
```
benfrancis, line 2865: Duplicate `id`

index.html
```diff
@@ -2736,64 +2821,80 @@ <h5><code>subscribeevent</code></h5>
...
+	  <p><span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-unsubscribeevent-1">
             In order to unsubscribe from an event, a Consumer MUST terminate 
             the corresponding Server-Sent Events connection with the Web Thing 
             as specified in the Server-Sent Events specification
             [[EVENTSOURCE]].
-          </p>
-          <p class="note" title="Terminating a connection using JavaScript">
+	  </span></p>
+	  <p><span class="note" title="Terminating a connection using JavaScript">
```
benfrancis, line 2882: Why the embedded <span>? Seems unnecessary.

index.html
```diff
@@ -2819,18 +2921,20 @@ <h5><code>subscribeallevents</code></h5>
...
+	  </span>
+	  <p><span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeallevents-3">
             The resolved value of the <code>href</code> member MUST then be used
             as the URL of the events resource.
+	  </span>
           </p>
-          <p class="rfc2119-assertion" 
+	  <p><span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeallevents-2">
```
benfrancis, line 2932: Duplicate `id`. Not the fault of this PR but easiest to fix here.

index.html
```diff
@@ -2862,14 +2966,14 @@ <h5><code>subscribeallevents</code></h5>
               const lampEventsSource = new EventSource('/things/lamp/events');
             </pre>
           </p>
-          <p class="rfc2119-assertion" 
+	  <p><span class="rfc2119-assertion" 
           id="http-sse-profile-protocol-binding-subscribeallevents-3">
```
benfrancis, line 2970: Duplicate `id`. Not the fault of this PR but easiest to fix here.

index.html
```diff
@@ -2885,49 +2989,64 @@ <h5><code>subscribeallevents</code></h5>
...
+	  <p><span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-unsubscribeallevents-1">
             In order to unsubscribe from all events, a Consumer MUST terminate 
             the corresponding Server-Sent Events connection with the events 
             endpoint of the Web Thing, following the steps specified in the 
             Server-Sent Events specification [[EVENTSOURCE]].
-          </p>
-          <p class="note" title="Terminating a connection using JavaScript">
+	  </span></p>
+	  <p><span class="note" title="Terminating a connection using JavaScript">
```
benfrancis, line 3049: Redundant `<span>`?

index.html
```diff
@@ -2736,64 +2821,80 @@ <h5><code>subscribeevent</code></h5>
...
               overheatedEventSource.close();
             </pre>
-          </p>
         </section>
   
         <section id="http-sse-profile-protocol-binding-events-subscribeallevents">
           <h5><code>subscribeallevents</code></h5>
-          <p class="rfc2119-assertion" 
+	  <p><span class="rfc2119-assertion" 
             id="http-sse-profile-protocol-binding-subscribeallevents-1">
```
benfrancis, line 2898: `http-sse-profile-protocol-binding-subscribeallevents-1` and `http-sse-profile-protocol-binding-subscribeallevents-2` need combining into a single assertion.

benfrancis (changes requested):
Thanks for doing this, sorry so many modifications were required. I will bear this in mind for future specification text.

I'm afraid there are quite a few instances of broken HTML that need fixing, plus some duplicate `id`s and two assertions that need combining into one.

FWIW I would personally suggest editing the assertion numbers to be sequential at this point since as far as I know nobody is currently relying on the existing numbering and we may as well start with a clean list.

I know that in other specifications the approach has been to sort the assertions alphabetically in an implementation report, but I note that some of the assertions in WoT Profile only currently make sense if read in sequence, since they describe sequential steps in a processing algorithm. E.g. "The resolved value of the href member MUST then be used as the URL of the Properties resource"

Indentation could be improved for readability, but that's true of the whole document.

Note: It's technically not valid HTML to put flow content like `<p>` and `<ul>` tags inside a `<span>` so it would be good if your tool could support `<div>`s (and semantic tags like `<p>`) in the future, but I can't really complain since I think I accidentally put `<ul>`s inside `<p>`s myself in some places!

testing/README.md
```diff
@@ -0,0 +1,219 @@
+# Implementation Report
+
+This directory contains the
+[WoT Thing Description Implementation report](https://w3c.github.io/wot-thing-description/testing/report.html)
```
mlagally, line 4: The documet needs to be updated for the profile. Links are still pointing to the TD.

testing/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Adding Your Results
+
+If you want to update your input for a Member organization, please go to [WoT IG testing](https://github.com/w3c/wot/tree/master/testing/tests/2019-05) and see the [README](https://github.com/w3c/wot/blob/master/testing/tests/2019-05/README.md) there.
+
+## Updating Implementation Report
+
+### w3c/wot Repository: Updating Results from Inputs
+
+Test inputs are collected via the WoT IG and are currently in `wot/testing/tests/2019-05`.
```
mlagally, line 19: This looks like an old link, please check.

testing/README.md
```diff
@@ -0,0 +1,219 @@
...
+[WoT Thing Description Implementation report](https://w3c.github.io/wot-thing-description/testing/report.html)
+and the sources for it.
+The report itself is in [report.html](report.html).
+Do not edit this file,
+it is autogenerated by executing `npm run assertions` from the main directory,
+and any edits would be overwritten the next time you generate it.
+
+## Adding Your Results
+
+If you want to update your input for a Member organization, please go to [WoT IG testing](https://github.com/w3c/wot/tree/master/testing/tests/2019-05) and see the [README](https://github.com/w3c/wot/blob/master/testing/tests/2019-05/README.md) there.
```
mlagally, line 13: 2019-05 looks to be outdated.

testing/README.md
```diff
@@ -0,0 +1,219 @@
...
+```
+and install the dependencies for the AssertionTester tool:
+```
+cd wot/testing/tools/thingweb-playground/AssertionTester
+npm install
+```
+
+---
+
+Go to the `wot/testing/tests/2019-05` directory.
```
mlagally, line 36: 2019?

testing/README.md
```diff
@@ -0,0 +1,219 @@
...
+is not JSON), this file should be empty.
+
+The output data files are now in `outputs/*/*.csv`.
+Check that there are no zero-length files in this list using `ls -l`.
+If there are it means something has gone wrong, most likely an input
+CSV file with a syntax error. An easy way to check for CSV syntax errors
+is to open the file in github. If it does not render properly as a table,
+there is an error, and GitHub will even point to the offending line.
+
+### w3c/wot-thing-description: Importing Results
```
mlagally, line 61: Thing Description

mlagally (changes requested):

mlagally:
The README.md still talks about TD, also the references to testing results point to 2019 directories.
Needs to be checked and updated for Profile.

To simplify review and merge, and to speed up the process, I suggest to separate the refactoring changes to index.html into a separate PR. Please also review Ben's comments.

index.html
```diff
@@ -894,6 +894,10 @@ <h3>Mandatory fields</h3>
             To provide minimum interoperability, the
             following metadata fields of a <a>Thing</a> MUST
             be contained in a compliant Thing Description:
+	    <em>
+	    FIX: this should have an table assertion per row, and this should be informative
+	    or a parent assertion.
+	    </em>
```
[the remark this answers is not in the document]
mmccool: Yeah, all rows should be assertions.  I have support for that in my tool, just need to do the markup.  There are lots of these in the document.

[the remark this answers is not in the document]
mmccool: BTW, when I was fixing this I noted that the assertion now labelled "profile-5.1.2.1-thing-mandatory-fields-1--id" says that the id must be "a globally unique urn of the thing".  So... we took OUT the word globally from the TD spec on purpose to resolve a privacy objection.  Basically the only way to get true global uniqueness is to have a central authority.  So anyway... not a good idea to try and put this back in via profiles, we'll just run into the same issue, again.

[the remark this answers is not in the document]
mmccool: Also... I'm not sure why security is listed here, it's already mandatory.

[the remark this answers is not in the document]
mmccool: anyway, fixed.  BTW the -- in the id indicates to my tool that these table assertions are sub-parts of the main assertion with the id before the --, so please don't modify these ids.  If we renumber them let's do in another pass to avoid confusion.  Also, descriptive names would be better than section references (which can easily become stale).

index.html
```diff
@@ -1053,6 +1057,10 @@ <h4 id="constraints-against-the-td-specification">Data
           <span class="rfc2119-assertion" id="profile-5.1.3-thing-data-schema-3">
           The following fields MUST be contained in a
           DataSchema:
+	  <em>
+	  FIX: this should have an table assertion per row, and this should be informative
+	  or a parent assertion.
+	  </em>
```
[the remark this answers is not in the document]
mmccool: fixed (removed, and table rows marked up with table assertions)

index.html
```diff
@@ -1281,17 +1296,18 @@ <h4>Recommended Practice</h4>
             or
             <code>bin</code>
             ,
-            <span class="rfc2119-assertion" id="profile-5.1.4.3-thing-properties-recommended-practice-2">
             to indicate how the value should be
-            interpreted. It is strongly RECOMMENDED to use
+            interpreted. 
+            <span class="rfc2119-assertion" id="profile-5.1.4.3-thing-properties-recommended-practice-2">
+	    It is strongly RECOMMENDED to use
```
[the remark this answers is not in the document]
mmccool: actually, put the span around the wrong sentence.  Moved.

index.html
```diff
@@ -1312,6 +1328,10 @@ <h4>Mandatory fields</h4>
             <span class="rfc2119-assertion" id="profile-5.1.5.1-thing-action-affordances-mandatory-fields-1">
             The following fields MUST be contained in an
             action element of an <a>Core Profile</a> compliant TD:
+	     <em>
+	     FIX: this should have an table assertion per row, and this should be informative
+	     or a parent assertion.
+	     </em>
```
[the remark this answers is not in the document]
mmccool: I had to restructure this to avoid having assertions nested inside assertions.  So now there is an assertion for each row (asserting the element must be present) and two SEPARATE assertions after the table indicating that these must use simple types.  There are some technical issues with these assertions though (dataSchema should be DataSchema?  What is a simple type?  If DataSchema can only have simple types, then why is objectSchema relevant?) but let's deal with those separately.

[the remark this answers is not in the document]
mmccool: Sigh... I'll have to update my tool to treat these correctly.  In the TD there was a column indicating whether a row element was mandatory or optional, but the implication here is that all of these are mandatory, and there is no such column.  A fairly simple tweak, though...

index.html
```diff
@@ -1377,6 +1397,10 @@ <h4>Additional Constraints</h4>
             applied to the Interaction Affordances of a <a>Thing
               Description</a> conforming to the <a>Core
               Data Model</a>:
+	     <em>
+	     FIX: this should have an table assertion per row, and this should be informative
+	     or a parent assertion.
+	     </em>
```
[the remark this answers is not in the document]
mmccool: This one is annoying because unlike other tables, it's not saying these elements are mandatory, it's actually a list of assertions.  Converting into a list of assertions, therefore.  The "_" is used to indicate these are Child assertions of the main assertion to keep the structure (although we could also just make the introductory paragraph informative).  Which DID require renumbering them since these were all 1, 2, 3 etc even though 2 was formerly a sub-part of 1, etc.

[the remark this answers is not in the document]
mmccool: I must say several of these are assertions which should have been added to TDs as bug fixes.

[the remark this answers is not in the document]
mmccool: to avoid getting into arguments about formatting, I left things as a table but use regular (child) assertions inside the table.  I did reword things a bit to make the assertions more self-contained and avoid starting sentences with lower case letters.

index.html
```diff
@@ -1574,6 +1598,10 @@ <h4>Mandatory fields</h4>
             <span class="rfc2119-assertion" id="profile-5.1.7.1-thing-forms-mandatory-fields-1">
             The following fields MUST be present in a form
             element of a <em>Core TD</em>:
+	  <em>
+	  FIX: this should have an table assertion per row, and this should be informative
+	  or a parent assertion.
+	  </em>
```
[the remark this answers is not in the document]
mmccool: this is a table indicating mandatory elements, so I used table assertions.

index.html
```diff
@@ -1615,6 +1643,10 @@ <h4>Additional Constraints</h4>
             applied to the Form elements of a <a>WoT Thing
               Description</a> conforming to the <a>Core
               profile</a>:
+	  <em>
+	  FIX: this should have an table assertion per row, and this should be informative
+	  or a parent assertion.
+	  </em>
```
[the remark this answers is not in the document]
mmccool: so... I'm converting this to the markup (it's a table which is a list of assertions as opposed to a set of mandatory fields), but just noticed it says "scopes MUST NOT be used".  Well... I kind of object to that, otherwise you are throwing away a valuable access control feature of OAuth, but I'll leave it for now and bring it up later.

index.html
```diff
@@ -1767,9 +1801,6 @@ <h5><code>writeproperty</code></h5>
             <a href="https://www.w3.org/TR/wot-thing-description/#propertyaffordance">
             <code>PropertyAffordance</code></a>
             for which:
-          </p>
-          <ul class="rfc2119-assertion" 
-            id="http-baseline-profile-protocol-binding-writeproperty-2">
```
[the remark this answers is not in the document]
mmccool: oops, and also the closing p tag.  fixed.

index.html
mmccool: ditto, fixed

index.html
mmccool: got two lines switched around.  fixed

index.html
mmccool: Ugh, this entire table is a combination of saying that things are optional or mandatory AND a set of constraints on their values, which is equivalent to multiple assertions.  I may have to restructure this completely and pull the constraint assertions out of the table.  Also now I can't assume that a table means that something is mandatory....

mmccool: although my comments about the table refer to invokeaction-12, where 2170 is now, so it looks like my previous edits have made line numbers unreliable.  Anyway, invokeaction-14 no longer exists, so... I assume you are referring to invokeaction-13, where I had a </ul> where there should have been a <ul>

index.html
mmccool: well, the problem is not so much automation but when an assertion is taken out of context, i.e. in the implementation report, it's much better to have complete, self-contained sentences.

mmccool: anyway, assume this is just a comment so I'll mark as resolved.

index.html
mmccool: ok, made the second one 7b

index.html
mmccool: sure, simplified

index.html
mmccool: sure, changed

index.html
mmccool: fixed, second one renumbered to 4

index.html
mmccool: sure, simplified

index.html
mmccool: fixed; a mistake in restructuring.

index.html
mmccool: fixed, I think

testing/README.md
mmccool: Actually, I updated this entire README at some point but it seems it did not get checked in, and now I can't find the version I edited.  The instructions here are now completely wrong.  Anyway, I edited the whole thing... again.

testing/README.md
mmccool: yeah, see above.   Whole file was obsolete.

testing/README.md
mmccool: whole file obsolete

testing/README.md
mmccool: again

mmccool:
I have addressed all of Ben's comments but noticed there were still a few assertions in tables that are not set up correctly, and also my tool needs some tweaks to deal with tables without a "mandatory" or "optional" column like in TDs (it turns each row into a MUST or MAY assertion).  The trouble is, there are also tables that are actually lists of assertions, and tables that are indicating BOTH that fields must or should be present AND adds constraints.  These are technically separate assertions from a testing point of view.

Regarding the issues about the README raised by Lagally, somehow an old version of that file survived; I had previously updated it but I guess forgot to commit it.  So I re-edited it.

I'm not going to waste time refactoring this, because as mentioned the changes to the markup are SO that I can regenerate the implementation report and as they are in separate files there are no conflicts.

There are, however, still some issues with the implementation report.  It would probably be a good idea to merge this PR and get caught up on some of the other pending PRs that it will cause conflicts with.  The remaining issues with the implementation report can be dealt with in another PR that does not have to touch index.html.

testing/README.md
mmccool: old file, this whole section is gone now (we are going to describe the data collection process in wot-testing, not in this readme)

PR #188, opened 2022-03-23
Split Core Profile into HTTP Baseline Profile and HTTP SSE Profile

benfrancis:
This PR proposes splitting the Core Profile out into an "HTTP Baseline Profile" and an "HTTP SSE" Profile, as discussed in #107.

This would allow for an "HTTP Webhooks Profile" to be defined separately which provides an alternative event mechanism for certain deployment scenarios where SSE doesn't scale as well (such as the one put forward by @mlagally).

Notes:
- The `observeproperty`, `unobserveproperty`, `observeallproperties` and `unobserveallproperties` protocol bindings currently targeted at the Core Profile in #157 would be added to the HTTP SSE Profile instead
- If we make this change we would also need to update the Abstract, Introduction and Terminology sections to replace instances of "Core Profile", and update appendix A
- I haven't touched the content of the Data Model section because that still requires a lot of changes since the clarification of specification requirements
- There are some parts of the HTTP Baseline Profile which might also be relevant for the HTTP SSE Profile and HTTP Webhooks Profile, such as Error Responses and Security.
- If people don't like the term "baseline" then alternative names for the HTTP Baseline Profile could be "HTTP REST Profile", or simply "HTTP Profile" (see #5).

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/188.html" title="Last updated on Mar 23, 2022, 4:59 PM UTC (1318c6b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/188/0eaf181...benfrancis:1318c6b.html" title="Last updated on Mar 23, 2022, 4:59 PM UTC (1318c6b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -693,7 +693,7 @@
     {
       "@context": "https://www.w3.org/2019/wot/td/v1",
       "id": "urn:dev:ops:32473-WoTLamp-1234",
-      "profile": "https://www.w3.org/2022/wot/profile/core/v1",
+      "profile": "https://www.w3.org/2022/wot/profile/http-baseline/v1",
       "title": "My Lamp",
       "description": "A web connected lamp",
       ...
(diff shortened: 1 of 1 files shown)


mlagally:
This looks great and brings the profile spec a big leap forward.

As you say, some additional factoring out of common error responses and security should be done, since these should be identical for all HTTP based profiles we define in this document.

mlagally (approved):
I also think the term "baseline profile" is appropriate, as it resolves the concers that were raised by some group members on the "core" name.

PR #191, opened 2022-04-08
Define protocol binding for observing properties - closes #95

### LINKED ISSUE #95: Should the Core Profile define a protocol binding for `observeproperty`/`unobserveproperty`?

benfrancis:
This PR replaces PR #157 and re-targets the SSE observe property protocol binding at the HTTP SSE profile, since the Core Profile has now been split. I have also restructured the assertion IDs to be consistent with #181.

This is a proposed protocol binding for observing properties in the HTTP SSE Profile. It includes bindings for the following WoT operations:
- `observeproperty`
- `unobserveproperty`
- `observeallproperties`
- `unobserveallproperties`

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/191.html" title="Last updated on May 11, 2022, 11:41 AM UTC (30370ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/191/1279a5f...benfrancis:30370ba.html" title="Last updated on May 11, 2022, 11:41 AM UTC (30370ba)">Diff</a>


benfrancis:
Replaced by #191.

mlagally (approved):

PR #195, opened 2022-04-29
Example TDs for HTTP Baseline and HTTP SSE profiles

benfrancis:
This PR replaces #91 and provides example Thing Descriptions used by the examples throughout the HTTP Baseline and HTTP SSE protocol binding sections.

Note that I have intentionally used a Thing Description which has both properties and events for the HTTP SSE Protocol Binding so that:
1. You can see how the HTTP SSE profile can build on top of the HTTP Baseline Profile
2. If and when #191 lands, I can modify the example to also include observing properties

I have not included the longer "declarative protocol binding illustration" example Thing Description from #91 because it no longer makes so much sense to have a single TD there, and decisions made for the Thing Description specification mean that parts of it are no longer valid anyway.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/195.html" title="Last updated on May 11, 2022, 8:36 AM UTC (131a0c8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/195/d41bd09...benfrancis:131a0c8.html" title="Last updated on May 11, 2022, 8:36 AM UTC (131a0c8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1685,6 +1685,83 @@
         MUST implement this protocol binding.
         </span>
       </p>
+
+      <p>
+        The examples provided throughout this section describe how a Consumer
+        would communicate with a Web Thing which produces the following Thing
+        Description:
+      </p>
+
+      <pre class="example">
+        {
+          "@context": "https://www.w3.org/2019/wot/td/v1",
+          "id": "https://mywebthingserver.com/things/lamp",
+          "profile": "https://www.w3.org/2022/wot/profile/http-baseline/v1",
+          "base": "https://mywebthingserver.com/things/lamp/",
+          "title": "My Lamp",
+          "description": "A web connected lamp",
+          "securityDefinitions": {
+            "oauth2": {
+              "scheme": "oauth2",
+              "flow": "code",
+              "authorization": "https://mywebthingserver.com/oauth/authorize",
+              "token": "https://mywebthingserver.com/oauth/token"
+            }
+          },
+          "security": "oauth2",
+          "properties": {
+            "on": {
+              "type": "boolean",
+              "title": "On/Off",
+              "description": "Whether the lamp is turned on",
+              "forms": [{"href": "./properties/on"}]
+            },
+            "level" : {
+              "type": "integer",
+              "title": "Brightness",
+              "description": "The level of light from 0-100",
+              "unit": "percent",
+              "minimum" : 0,
+              "maximum" : 100,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1685,6 +1685,83 @@ <h3>Protocol Binding</h3>
...
+                    "unit": "percent"
+                  },
+                  "duration": {
+                    "type": "integer",
+                    "minimum": 0,
+                    "unit": "milliseconds"
+                  }
+                }
+              },
+              "forms": [{"href": "./actions/fade"}]
```
mlagally, line 1749: why "./" prefix?

index.html
```diff
@@ -2627,6 +2704,102 @@ <h2>Protocol Binding</h2>
...
+            }
+          },
+          "events": {
+            "overheated": {
+              "title": "Overheated",
+              "data": {
+                "type": "number",
+                "unit": "degree celsius"
```
mlagally, line 2777: please consider using a well known unit identifier, e.g. from QUDT.
Need to use a QUDT context file.
How to get a renderable unit? QUDT lookup or embed in the TD?

index.html
benfrancis: I added the prefix for the top level "properties", "actions" and "events" endpoints to make it clearer that the `href` member is a path, then the rest are just there for consistency. I've removed the prefixes now though since apparently they are confusing.

index.html
benfrancis: I think this requires further discussion in #29 as to whether the profile mandates a particular unit ontology.

In WebThings we use the long form of [SI units](https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf/2d2b50bf-f2b4-9661-f402-5f9d66e4b507?version=1.10&t=1632138671693&download=true) for the values of `unit` members, e.g. "volt", "ohm", "lux", "lumen", "hertz", "watt" etc.

mlagally (approved):
This is a very useful contribution which should be merged.
At a later time we may nead to adjust the example a bit when some open questions have been settled.

index.html
mlagally: Arch call on May 11th:
Agree to merge this PR, discussion should continue in #29 - We could use SI units and potentially also permit QUDT with a prefix. Everything could be converted to SI units.

PR #198, opened 2022-05-05
WIP: event payload format

mlagally:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/198.html" title="Last updated on Jun 28, 2022, 6:02 PM UTC (7346066)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/198/b0ebae6...7346066.html" title="Last updated on Jun 28, 2022, 6:02 PM UTC (7346066)">Diff</a>


index.html
```diff
@@ -3053,15 +3053,105 @@ <h2>Identifier</h2>
...
+                A WebThing MAY resend the same event with the same <code>id</code>. 
+                If an event is resent with the same <code>id</code>, a consumer SHOULD treat it as
+                duplicate.</td>
+            </tr>
+            <tr>
+              <td>source</td>
+              <td>uri</td>
+              <td>mandatory, non-empty</td>
+              <td>The combination of <code>source</code> + <code>id</code> MUST be unique for each distinct event. 
+                The source MUST be set to the <code>baseURI</code> of the WebThing.</td>
```
egekorkan, line 3112: what is `baseURI`? Why do we need this if we have TDs?

index.html
```diff
@@ -3053,15 +3053,105 @@ <h2>Identifier</h2>
...
+            <tr>
+              <td>data</td>
+              <td>string</td>
+              <td>optional</td>
+              <td>JSON encoded <code>data</code> object of the event affordance.</td>
+            </tr>
+          </tbody>
+        </table>
+        <span class="rfc2119-assertion" id="profile-event-payload-size">
+          The size of an event object MUST NOT exceed 64KB, to ensure interoperability with all kinds of consumers.
```
egekorkan, line 3153: where does this come from?

egekorkan (commented):
Even though it is a WIP, I had a look. I find a fundamental problem with using cloud events in a profile that is meant for greenfield applications. It adds too much metadata to payload and does not make sense in a WoT specification.

index.html
```diff
@@ -3053,15 +3053,105 @@ <h2>Identifier</h2>
...
+            </tr>
+            <tr>
+              <td>time</td>
+              <td>string</td>
+              <td>optional, compliant with [[RFC3339]]</td>
+              <td>time when the event occurred. If the WebThing does not have a clock, it MUST use the same
+                algorithm on all event affordances to determine the value of this field.
+            </tr>
+            <tr>
+              <td>datacontenttype</td>
```
sebastiankb, line 3139: why do we need 'datacontenttype'. forms already have a 'contentType' defined

index.html
```diff
@@ -3053,15 +3053,105 @@ <h2>Identifier</h2>
...
+              <td>string</td>
+              <td>optional, [[RFC2046]] compliant MIME type, when set the value MUST be "<code>application/json</code>"</td>
+              <td>If a WebThing contains a <code>data member</code>, this field MUST be present.</td>
+            </tr>
+            <tr>
+              <td>data</td>
+              <td>string</td>
```
sebastiankb, line 3146: why do we have a `string` here?

index.html
```diff
@@ -3053,15 +3053,105 @@ <h2>Identifier</h2>
...
+          <tbody>
+            <tr>
+              <td>specversion</td>
+              <td>string</td>
+              <td>mandatory, non-empty</td>
+              <td>The <code>specversion</code> MUST have the value "1.0"</td>
+            </tr>
+            <tr>
+              <td>id</td>
+              <td>string</td>
```
sebastiankb, line 3100: TD uses anyURI

index.html
mlagally: A pragmatic limit. 
Why would we have more than 64k for an event payload?

We should discuss the limits together.

egekorkan: Similarly, why would need more than 32KB? Or 35KB?

index.html
mlagally: It is a URI in the original sense, i.e. an indentifier for a unique resource (= the thing)

index.html
mlagally: 64K can be addressed by 16 bits, which is still a limit on some platforms.

index.html
mlagally: good point - we could restrict the string contents to URI formats only.

mlagally:
Won't merge - superseeded by PR 235

index.html
sebastiankb: image

I think, it should be definitely any type.

PR #207, opened 2022-05-25
updating JSON Schema Validation reference

mlagally:


mlagally:
@sebastiankb , @egekorkan , @benfrancis - can you please review and check consistency with TD spec.

egekorkan:
I propose to close this PR without merging it. The updated version is draft 2020-12 and the current one is draft 7 that the TD spec is using. We have decide to stick with draft 7 until version 2. If in doubt, the IETF drafts have a section called meta schema which contains the draft version. See the links below:
- https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-5
- https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-5

PR #208, opened 2022-05-25
adding missing image

mlagally:


benfrancis:
Note that the unsubscribe operation in this diagram uses a `DELETE` request, but in the protocol binding it was changed to a `POST` request.

My instinct is that a `DELETE` request on a subscription URL would be better, but maybe there were reasons for choosing `POST`.

benfrancis:
Since this PR was already merged, I've filed #254 to track this.

PR #236, opened 2022-07-04
image update: rename core -> baseline

mlagally:


egekorkan (commented):
Not approving or requesting changes but shouldn't we clean up this WoT Enhanced profile? It feels odd at this point, given that we have 3 profiles

mlagally:
@egekorkan I agree that we have to update the figure, I filed https://github.com/w3c/wot-profile/issues/237
for that task.

This MR is just the name change - I suggest we merge it and do the update üer #237 afterwards.

PR #257, opened 2022-08-03
Assert that the profiling mechanism depends on TD v1.1 - closes #217

### LINKED ISSUE #217: TD version dependency
The spec is not clear about which version of the TD spec must be used. Example 3 uses 1.0 even though the terms that are required by the HTTP Baseline Profile such as `queryaction` are defined only in the TD spec V1.1 . 

I think that the profile spec should be very strict on this and say that only 1.1 TDs can be used.

benfrancis:
As per the suggestion in issue #217, this PR adds a new assertion to make it clear that the profiling mechanism depends on Thing Description version 1.1:

> In order to use a profile member in a Thing Description, the `@context` member MUST contain the anyURI https://www.w3.org/2022/wot/td/v1.1 in order to denote that the document is using version 1.1 of the Thing Description specification. [[wot-thing-description11](file:///home/tola/Code/benfrancis/wot-profile/index.html#bib-wot-thing-description11)].

I've updated all the Thing Description examples to conform with this assertion.

I've also added missing RFC 2119 markup to all the assertions in the Profiling Mechanism section.


mlagally (approved):
Thanks @benfrancis for this improved text.

egekorkan (approved):

egekorkan (approved):

benfrancis:
@mlagally I have resolved the merge conflicts and I believe this is now ready to merge.

PR #260, opened 2022-08-10
Re-structure profiles sections - closes #214 and closes #248 

benfrancis:
This is a proposed re-structuring of the profiles sections, based on the discussion in #214.

Overview of changes:
1. Renamed "HTTP Profile" to "Common Constraints" 
2. Moved Default Language, Links and Security Schemes sections to the top level of this section
3. Moved Information Model, Discovery and Errors sections from HTTP Baseline Profile to Common Constraints
4. Merged the Security section from HTTP Baseline Profile into the Security section of Common Constraints and removed the placeholder Transport Security section in favour of the note referring to the Security Best Practices document
5. Added a placeholder section for Semantic Annotations in Common Constraints
7. Moved Privacy Considerations and Security Considerations into their own top level sections at the end of the document
8. Added assertions to the three profiles to say that they require the constraints in the Common Constraints section to be conformed with
9. Renamed a bunch of HTML element IDs to reflect the new structure
10. Re-worded some assertions to reflect renamed section titles

After these changes, at a high level the specification looks like:
- Common Constraints
  - Information Model
  - Security
  - Discovery
  - Links
  - Errors
  - Semantic Annotations
  - Default Language
- HTTP Baseline Profile
  - Identifier
  - Protocol Binding
- HTTP SSE Profile
  - Identifier
  - Protocol Binding
- HTTP Webhook Profile
  - Introduction
  - Identifier
  - Message Format
  - Protocol Binding
- Privacy Considerations
- Security Considerations

If we land these changes then we'll also need to update the Introduction and Abstract to reflect renamed sections. I didn't do that in this PR because it's already large enough!

I also think we should consider renaming "HTTP Baseline Profile" to "HTTP Basic Profile", since the name "Baseline Profile" may now be confused with "Common Constraints".

My comments in https://github.com/w3c/wot-profile/issues/214#issuecomment-1203761916 regarding the Information Model and Message Format sections are left unresolved for now.

Note: This diff is hard to read because it touches so many lines, but it's slightly easier if you just view the second commit (the first one is pulled in from #249 and may be re-based out if that lands first).

Edit: I've renamed HTTP Profile Common Constraints to Common Constraints for simplicity, since all the profiles are currently HTTP profiles. We can re-visit if other profiles are added in the future.

***
[...]

PR #264, opened 2022-08-24
Add geolocation semantic annotation recommendation - closes #137

### LINKED ISSUE #137: Geolocation

benfrancis:
This assertion recommends the use of the schema.org ontology for geolocation semantic annotations, as discussed in #17 and #137.

I've included two examples of how to use the geolocation annotations which I think are equally valid:
1. To add static geolocation metadata to a Thing at the top level of its Thing Description
2. To add a dynamic geolocation property to a Thing as a PropertyAffordance

Note that this PR depends on #260 which adds a Semantic Annotations section under Common Constraints. Please only review the final commit https://github.com/w3c/wot-profile/commit/cfa5716c9809364647d55b6c1eb1fea564a6fd2f

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/264.html" title="Last updated on Aug 31, 2022, 11:50 AM UTC (6bc99ec)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/264/68e06a2...benfrancis:6bc99ec.html" title="Last updated on Aug 31, 2022, 11:50 AM UTC (6bc99ec)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1831,12 +1831,56 @@
 
     <section id="common-constraints-semantic-annotations">
       <h2>Semantic Annotations</h2>
-      <p class="ednote">
-        TODO: Define a set of recommended semantic ontologies for common
-        vocabulary such as
-        <a href="https://github.com/w3c/wot-profile/issues/137">geolocation</a>
-        and <a href="https://github.com/w3c/wot-profile/issues/29">units</a>.
-      </p>
+      <section id="common-constraints-semantic-annotations-geolocation">
+        <h3>Geolocation</h3>
+        <p>
+          <span class="rfc2119-assertion" id="common-constraints-errors-7">
+            It is RECOMMENDED that where geolocation metadata is needed, the 
+            schema.org [[schema-org]] ontology is used.
+          </span>
+        </p>
+        <pre class="example" title="Geolocation annotations on a Thing">
+          {
+            "@context": [
+              "https://www.w3.org/2022/wot/td/v1.1",
+              {
+                "schema": "https://schema.org"
+              }
+            ],
+            "schema:longitude": "297.83",
+            "schema:latitude": "26.58",
+            "schema:elevation": "5.3"
+          }
+        </pre>
+        <pre class="example" title="Geolocation annotations on a PropertyAffordance">
+          {
+            "@context": [
+              "https://www.w3.org/2022/wot/td/v1.1",
+              {
+                "schema": "https://schema.org"
+              }
+            ],
+            "properties": {
+              "position": {
+                "type": "object",
+                "@type": "schema:GeoCoordinates",
+                "properties": {
+                  "longitude": { "type": "number" },
+                  "latitude":  { "type": "number" },
+                  "elevation": { "type": "number" }
+                }
+              },
+            },
+          }
+        </pre>
+      </section>
+      <section id="common-constraints-semantic-annotations-units">
+        <h2>Units</h2>
+        <p class="ednote">
+          TODO: Define a recommended semantic ontology for 
+          <a href="https://github.com/w3c/wot-profile/issues/29">units</a>.
+        </p>
+      </section>
     </section>
 
     <section id="sec-default-language">


mlagally (changes requested):
This MR contains changes out of scope of the geolocation section.

I reviewed that section, i.e. lines 1834-1884 only.
We have to be more specific on which of the schema.org properties should be used, to define which types from schema.org MUST be supported on the consumer. 

I think we should clarify that only properties for GeoCoordinates:elevation, altitude and longitude are defined in the profile.
On elevation there's a dependency of 'UNITOFMEASUREMENT' with a default of meters. We should clarify if non-metric values must be supported, or if we constrain to meters and leave it to the consumer to convert to the locale of the user.



Properties from GeoCoordinates
--
PostalAddress  or Text | Physical address of the item.
Country  or Text | The country. For example, USA. You can also provide the two-letter ISO 3166-1 alpha-2 country code.
Number  or Text | The elevation of a location (WGS 84). Values may be of the form 'NUMBER UNITOFMEASUREMENT' (e.g., '1,000 m', '3,200 ft') while numbers alone should be assumed to be a value in meters.
Number  or Text | The latitude of a location. For example 37.42242 (WGS 84).
Number  or Text | The longitude of a location. For example -122.08585 (WGS 84).
Text | The postal code. For example, 94043.

Properties from [GeoCoordinates](https://schema.org/GeoCoordinates)
[address](https://schema.org/address)	[PostalAddress](https://schema.org/PostalAddress)  or 
[Text](https://schema.org/Text)	Physical address of the item.
[addressCountry](https://schema.org/addressCountry)	[Country](https://schema.org/Country)  or 
[Text](https://schema.org/Text)	The country. For example, USA. You can also provide the two-letter [ISO 3166-1 alpha-2 country code](http://en.wikipedia.org/wiki/ISO_3166-1).
[elevation](https://schema.org/elevation)	[Number](https://schema.org/Number)  or 
[Text](https://schema.org/Text)	The elevation of a location ([WGS 84](https://en.wikipedia.org/wiki/World_Geodetic_System)). Values may be of the form 'NUMBER UNITOFMEASUREMENT' (e.g., '1,000 m', '3,200 ft') while numbers alone should be assumed to be a value in meters.
[latitude](https://schema.org/latitude)	[Number](https://schema.org/Number)  or 
[Text](https://schema.org/Text)	The latitude of a location. For example 37.42242 ([WGS 84](https://en.wikipedia.org/wiki/World_Geodetic_System)).
[longitude](https://schema.org/longitude)	[Number](https://schema.org/Number)  or 
[Text](https://schema.org/Text)	The longitude of a location. For example -122.08585 ([WGS 84](https://en.wikipedia.org/wiki/World_Geodetic_System)).
[postalCode](https://schema.org/postalCode)	[Text](https://schema.org/Text)	The postal code. For example, 94043.

benfrancis:
> This MR contains changes out of scope of the geolocation section.

Please see my comment from above:

>> Note that this PR depends on https://github.com/w3c/wot-profile/pull/260 which adds a Semantic Annotations section under Common Constraints. Please only review the final commit https://github.com/w3c/wot-profile/commit/cfa5716c9809364647d55b6c1eb1fea564a6fd2f

This text is being added to a section created in #260 and so depends on that PR. I can rebase this PR if and when that one lands.

> We have to be more specific on which of the schema.org properties should be used, to define which types from schema.org MUST be supported on the consumer.

I don't object to the idea of making support for schema.org's `latitude`, `longitude` and `elevation` properties mandatory, if we can define what "supports" means, to the extent that the assertions are testable.

However, I think we need to have a group discussion about the principle of making support for for certain semantic annotations mandatory for Consumers. Semantic annotations are intended as hints to the Consumer about additional semantic context of data emitted from or consumed by a Thing, but it's difficult to define exactly what a Consumer needs to implement in order to "support" them. What does it mean for a Consumer to "support" geolocation annotations? Does it need to be able to plot them on a map, plot them on a 3D model, import them into a knowledge graph, or simply display the values to a user? What if the Consumer doesn't have a graphical user interface?

Additionally, if we accept the principle that some semantic annotations must be supported by Consumers in all three of the current profiles, then it opens the door to expand this to many other semantic annotations and ontologies in the future. What criteria do we use to decide which annotations should be mandatory to support, and where do we stop?

Simply _recommending_ support for certain very common cross-domain semantic annotations could be a compromise which helps improve interoperability by nudging implementations towards certain ontologies, whilst not creating too much of an implementation burden for Consumers which may not actually need them, and accepting the richness that the extensible nature of semantic extensions allows.

index.html
```diff
@@ -1831,12 +1831,56 @@ <h2>Errors</h2>
...
-      <p class="ednote">
-        TODO: Define a set of recommended semantic ontologies for common
-        vocabulary such as
-        <a href="https://github.com/w3c/wot-profile/issues/137">geolocation</a>
-        and <a href="https://github.com/w3c/wot-profile/issues/29">units</a>.
-      </p>
+      <section id="common-constraints-semantic-annotations-geolocation">
+        <h3>Geolocation</h3>
+        <p>
+          <span class="rfc2119-assertion" id="common-constraints-errors-7">
```
sebastiankb, line 1837: I think, it is not a good idea to concentrate only on a single geo ontology. The proposal should be open to use different geo location approaches that, e.g., work indoor and outdoor. Building typical uses different location information such as zone, level, room number, etc. . I think many or most IoT devices are used within buildings, so we should support these approaches.

sebastiankb (approved):
I would not support this PR since geolaction has different facets. There are different approaches

benfrancis:
Closing, as we will re-visit this in 1.1. Tagged as Profile-1.1 to remind us.

PR #266, opened 2022-08-24
Refine sync vs. async action protocol binding - closes #259

### LINKED ISSUE #259: Is it OK to not have `output` specified in a profile TD
The examples (e.g,. [Example 3](https://w3c.github.io/wot-profile/#example-3)) describe actions without describing  `output` .

```
  "actions": {
    "fade": {
      "title": "Fade",
      "description": "Fade the lamp to a given level",
      "input": {
        "type": "object",
        "properties": {
          "level": {
            "type": "integer",
            "minimum": 0,
            "maximum": 100,
            "unit": "percent"
          },
          "duration": {
            "type": "integer",
            "minimum": 0,
            "unit": "milliseconds"
          }
        }
      },
      "forms": [{"href": "actions/fade"}]
    }
  }
```

The result of an action in the Profile is **always** [ActionStatus](https://w3c.github.io/wot-profile/#ActionStatus). The Profile TD does not describe that situation using  `output` .
I understand the idea to reduce TD verbosity but wouldn't it be better to have at least

```
      "output": {
        "type": "object"
      }
```

A generic implementation of a WoT runtime otherwise would not expect any result.

Side-note: It may depend on the interpretation but [7.2.5.1 Mandatory fields](https://w3c.github.io/wot-profile/#mandatory-fields-1) mentions only title being mandatory. Is this correct? Don't we always need output ... even if being implicit?

benfrancis:
This is a WIP PR to address the issues raised in #259.

Summary of changes:
- Change the synchronous action protocol binding to return only the output of the action in the body of its response (instead of the `ActionStatus` object used by async responses), so that the response matches the output data schema of the ActionAffordance. This means that even non-conformant Consumers will be able to understand the output of synchronous actions.
- Assert that whether a Web Thing responds with a sync or async response depends on the `synchronous` member of the ActionAffordance

If we decide to adopt this change to the protocol binding then I think we'll need to change the examples to include an action which has an output. Currently the example Thing used throughout the protocol binding only has an action which has an input but no output. To be honest I'm struggling to think of an example of an action which has both an input and and output.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/266.html" title="Last updated on Sep 28, 2022, 11:19 AM UTC (d9a65b0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/266/3d4a8cb...benfrancis:d9a65b0.html" title="Last updated on Sep 28, 2022, 11:19 AM UTC (d9a65b0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1187,7 +1187,7 @@
               <li>URL set to the URL of the <code>Property</code> resource</li>
               <li><code>Accept</code> header set to <code>application/json
                 </code></li>
-            </ul>
+            </ul> 
           </div>
           <pre class="example">
               GET /things/lamp/properties/on HTTP/1.1
@@ -1506,7 +1506,38 @@
               </li>
             </ol>
           </div>
-          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-8">
+          <p>
+            <span class="rfc2119-assertion" 
+            id="http-basic-profile-protocol-binding-invokeaction-7">
+              If the <code>synchronous</code> member of the 
+              <a href="https://w3c.github.io/wot-thing-description/#actionaffordance"><code>ActionAffordance</code></a>
+              [[wot-thing-description11]] is set to <code>true</code> then the Web 
+              Thing MUST respond with a
+              <a href="#sync-action-response">Synchronous Action Response</a>.
+            </span>
+          </p>
+          <p>
+            <span class="rfc2119-assertion" 
+            id="http-basic-profile-protocol-binding-invokeaction-8">
+              If the <code>synchronous</code> member of the 
+              <a href="https://w3c.github.io/wot-thing-description/#actionaffordance"><code>ActionAffordance</code></a>
+              [[wot-thing-description11]] is set to <code>false</code> then the Web 
+              Thing MUST respond with an
+              <a href="#async-action-response">Asynchronous Action Response</a>.
+            </span>
+          </p>
+          <p>
+            <span class="rfc2119-assertion" 
+            id="http-basic-profile-protocol-binding-invokeaction-9">
+              If the <code>synchronous</code> member of the 
+              <a href="https://w3c.github.io/wot-thing-description/#actionaffordance"><code>ActionAffordance</code></a>
+              [[wot-thing-description11]] is <code>undefined</code> then the Web 
+              Thing MAY respond with either a 
+              <a href="#sync-action-response">Synchronous Action Response</a> or
+              <a href="#async-action-response">Asynchronous Action Response</a>.
+            </span>
+          </p>
+          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-10">
               For long-running actions which are not expected to finish executing
               within the timeout period of an HTTP request (e.g. 30 to 120
               seconds), it is RECOMMENDED that a Web Thing respond with an
@@ -1515,28 +1546,28 @@
               <code>queryaction</code> operation on a dynamically created
               <code>ActionStatus</code> resource, after the initial
               <code>invokeaction</code> response.</span></p>
-          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-9">
+          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-11">
               For short-lived actions which are expected to finish executing
               within the timeout period of an HTTP request, a Web Thing MAY wait
               until the action has completed to send a Synchronous Action
               Response.</span></p>
-          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-10">
+          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-12">
               If a Web Thing encounters an error in attempting to execute an
               action before responding to the <code>invokeaction</code> request,
               then it MUST send an Error Response.</span></p>
-          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-11a">
+          <p><span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-13">
               Conforming Consumers MUST support all three types of response
               to the initial <code>invokeaction</code> request.</span>
-            <span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-11b">
+            <span class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-14">
               After the initial request,
               support for subsequent operations on an <code>ActionStatus</code>
               resource is OPTIONAL.</span>
           </p>
 
           <h6 id="ActionStatus"><code>ActionStatus</code> object</h6>
-          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-12">
+          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-15">
             <p>
-              The status of an action invocation request is represented by an
+              The status of an asynchronous action invocation request is represented by an
               <code>ActionStatus</code> object which includes the following
               members:
             </p>
@@ -1550,7 +1581,7 @@
                 </tr>
               </thead>
               <tbody>
-                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-12_1">
+                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-16">
                   <td><code>status</code></td>
                   <td>The status of the action request.</td>
                   <td>mandatory</td>
@@ -1561,7 +1592,7 @@
                     <code>failed</code>)
                   </td>
                 </tr>
-                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-12_2">
+                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-17">
                   <td><code>output</code></td>
                   <td>
                     The output data, if any, of a completed action which
@@ -1573,7 +1604,7 @@
                   <td>optional</td>
                   <td>any type</td>
                 </tr>
-                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-12_3">
+                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-18">
                   <td><code>error</code></td>
                   <td>
                     An error message, if any, associated with a failed action
@@ -1585,7 +1616,7 @@
                   <td>optional</td>
                   <td><code>object</code></td>
                 </tr>
-                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-12_4">
+                <tr class="rfc2119-table-assertion" id="http-basic-profile-protocol-binding-invokeaction-19">
                   <td><code>href</code></td>
                   <td>
                     The [[URL]] of an <code>ActionStatus</code> resource which can
@@ -1626,7 +1657,6 @@
               </tbody>
             </table>
           </div>
-
           <p class="note">
             It is possible that a Thing's clock may not be set to the correct
             time. If timings are important then a Consumer may therefore choose
@@ -1637,7 +1667,7 @@
           </p>
 
           <h6 id="sync-action-response">Synchronous Action Response</h6>
-          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-13">
+          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-20">
             <p>
               If providing a Synchronous Action Response, a Web Thing MUST send
               an HTTP response with:
@@ -1647,23 +1677,17 @@
               <li><code>Content-Type</code> header set to
                 <code>application/json</code>
               </li>
-              <li>A body containing an <a href="#ActionStatus">
-                  <code>ActionStatus</code></a> object serialized
+              <li>A body containing the output of the action, if any, serialized
                 in JSON</li>
             </ul>
           </div>
           <pre class="example">
             HTTP/1.1 200 OK
             Content-Type: application/json
-            {
-              "status": "completed",
-              "timeRequested": "2021-11-10T11:43:20.135Z",
-              "timeEnded": "2021-11-10T11:43:25.135Z"
-            }
           </pre>
 
           <h6 id="async-action-response">Asynchronous Action Response</h6>
-          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-15">
+          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-21">
             <p>
               If providing an Asynchronous Action Response, a Web Thing MUST send
               an HTTP response containing the URL of an <code>ActionStatus</code>


mlagally:
@benfrancis 
Many thanks for drafting this proposal.

If the  synchronous  member of the [ ActionAffordance ] is undefined  then the Web Thing  MAY  respond with either a [ Synchronous Action Response ] or [ Asynchronous Action Response ].

The first two alternatives (synchronous = true/false) are well-defined, however the ambiguity in the "undefined" case is potentially dangerous, so the profile should make this flag mandatory.

benfrancis:
@mlagally wrote:
> The first two alternatives (synchronous = true/false) are well-defined, however the ambiguity in the "undefined" case is potentially dangerous, so the profile should make this flag mandatory.

This proposed default is actually the case for all Web Things with the current specification - a Web Thing can respond with either a synchronous or asynchronous response. I personally don't think it's a big problem because the two response types are clearly distinguished by the `200` vs. `201` response code.

What do others think? Should the `synchronous` member be mandatory for all `ActionAffordance`s of a Web Thing compliant with the HTTP Baseline Profile? Regardless of what the default is, Consumers have to support both response types, so it's really a question of verbosity in the Thing Description. I would personally leave it as optional. @sebastiankb @danielpeintner

danielpeintner:
I have to admit, that in the first place, I am still not super happy that `output` is not _always_ properly describing the result of an action but instead depends on `synchronous` term also:
* `sync` -> `output` type matches with the result of an action while
* `async` -> `output` type is wrapped into an `ActionStatus` object

benfrancis:
@danielpeintner The output schema does always describe the output of the action, but the output can not always be provided in the HTTP response to an invokeaction request, because some actions in the real world take longer than the HTTP request timeout period. This is a limitation of the underlying protocol, and exactly the problem the action queue pattern is designed to solve.

See also: https://github.com/w3c/wot-thing-description/issues/1665. I think that the response not containing the output is exactly what synchronous = false implies.

Do you have an alternative proposed solution for an HTTP protocol binding for asynchronous action requests? Unfortunately despite it being one of the goals for the current charter, there is still no obvious way to describe an action queue declaratively in a Thing Description.

benfrancis:
> I prefer that output always describes the result of an action.

By "result of an action" I assume you mean the response to the `invokeaction` request, not the actual output of the action (which may not have completed at the time the response arrives).

I think you can see why that would result in extremely verbose Thing Descriptions which whilst technically correct aren't actually any more useful to the Consumer in practice (since the Consumer won't understand what to do with the ActionStatus object).

It also breaks down in cases where there are multiple protocol bindings for the same action affordance which have different payload formats. For example, WebThings Gateway provides both an HTTP and WebSockets API for invoking actions. If the `output` member of the data schema is used to describe the payload format of the HTTP API (i.e. the protocol binding), that would not be valid for use in the WebSocket sub-protocol, which necessarily requires a different payload format due to the difference in semantics of the underlying protocol.

> I know.. this generic implementation would not be able to properly parse (or understand) Action Status but at least it would not try to cast ActionStatus to the output type and throw an error that the result does not match output. Do you know what I mean?

Yes, I agree that the `ActionStatus` response would cause problems for Consumers which don't implement the profile. But I would argue that this is due to a limitation of Thing Descriptions which will apply to all asynchronous action affordances. There is currently no way to provide a separate data schema for the output of an action and the response to an invokeaction request, where they are different. There is an implicit conflation of the data schema of the action output and its protocol binding.

I think the solution to this problem would be to provide a [schema member for `ExpectedResponse`](https://github.com/w3c/wot-thing-description/issues/1053). That way you can cleanly separate the data schema of the actual output of the real world action (`output`) from the data schema of the response to the invokeaction operation in the protocol binding (`response`).

In the meantime, if you really want to describe the HTTP Baseline Profile's protocol binding declaratively in a Thing Description (which I would argue is not necessary), perhaps an alternative is to use `AdditionalResponse` instead of `ExpectedResponse`?

Of course, how to describe the protocol bindings for `queryaction`, `cancelaction` and `queryallactions` is a whole other matter. Perhaps you can improve on my attempt?

```json
{
  "@context": "https://www.w3.org/2019/wot/td/v1",
  "id": "https://mywebthingserver.com/things/lamp",
  "base": "https://mywebthingserver.com/things/lamp/",
  "title": "My Lamp",
  "description": "A web connected lamp",
  "securityDefinitions": {
    "oauth2": {
      "scheme": "oauth2",
      "flow": "code",
      "authorization": "https://mywebthingserver.com/oauth/authorize",
      "token": "https://mywebthingserver.com/oauth/token"
    }
  },
  "security": "oauth2",
  "properties": {
    "on": {
      "type": "boolean",
      "title": "On/Off",
      "description": "Whether the lamp is turned on",
      "forms": [
        {
          "op": "readproperty",
          "href": "./properties/on",
          "htv:methodName": "GET",
          "htv:headers": [{
            "htv:fieldName": "Accept",
            "htv:fieldValue": "application/json"
          }],
          "response": {
            "htv:statusCodeNumber": 200,
            "contentType": "application/json"
          }
        },
        {
          "op": "writeproperty",
          "href": "./properties/on",
          "htv:methodName": "PUT",
          "contentType": "application/json",
          "response": {
            "htv:statusCodeNumber": 204
          }
        }
      ]
    },
    "level" : {
      "type": "integer",
      "title": "Brightness",
      "description": "The level of light from 0-100",
      "unit": "percent",
      "minimum" : 0,
      "maximum" : 100,
      "forms": [
        {
          "op": "readproperty",
          "href": "./properties/level",
          "htv:methodName": "GET",
          "htv:headers": [{
            "htv:fieldName": "Accept",
            "htv:fieldValue": "application/json"
          }],
          "response": {
            "htv:statusCodeNumber": 200,
            "contentType": "application/json"
          }
        },
        {
          "op": "writeproperty",
          "href": "./properties/level",
          "htv:methodName": "PUT",
          "contentType": "application/json",
          "response": {
            "htv:statusCodeNumber": 204
          }
        }
      ]
    }
  },
  "actions": {
    "fade": {
      "title": "Fade",
      "description": "Fade the lamp to a given level",
      "input": {
        "type": "object",
        "properties": {
          "level": {
            "type": "integer",
            "minimum": 0,
            "maximum": 100,
            "unit": "percent"
          },
          "duration": {
            "type": "integer",
            "minimum": 0,
            "unit": "milliseconds"
          }
        }
      },
      "forms": [
        {
          "op": "invokeaction",
          "href": "./actions/fade",
          "htv:methodName": "POST",
          "contentType": "application/json",
          "htv:headers": [{
            "htv:fieldName": "Accept",
            "htv:fieldValue": "application/json"
          }],
          "response": {
            "htv:statusCodeNumber": 201,
            "htv:headers": [
              {
                "htv:fieldName": "Location",
                "htv:fieldValue": "/fade/{id}"
              }
            ],
            "contentType": "application/json",
            "schema": "actionStatus"
          },
          "additionalResponses": [
            {
              "htv:statusCodeNumber": 200,
              "htv:headers": [
                {
                  "htv:fieldName": "Location",
                  "htv:fieldValue": "/fade/{id}"
                }
              ],
              "contentType": "application/json",
              "schema": "actionStatus"
            }
          ]
        },
        {
          "op": "queryaction",
          "href": "./actions/fade/{id}",
          "htv:methodName": "GET",
          "htv:headers": [{
            "htv:fieldName": "Accept",
            "htv:fieldValue": "application/json"
          }],
          "response": {
            "contentType": "application/json",
            "schema": "actionStatus"
          }
        },
        {
          "op": "cancelaction",
          "href": "./actions/fade/{id}",
          "htv:methodName": "DELETE",
          "contentType": "application/json"
        }
      ],
      "uriVariables": {
        "id": {
          "type": "string",
          "description": "identifier of action request"
        }
      }
    }
  },
  "events": {
    "overheated": {
      "title": "Overheated",
      "data": {
        "type": "number",
        "unit": "degree celsius"
      },
      "description": "The lamp has exceeded its safe operating temperature",
      "forms": [
        {
          "op": "subscribeevent",
          "href": "./events/overheated",
          "subprotocol": "sse",
          "htv:methodName": "GET",
          "htv:headers": [
            {
              "htv:fieldName": "Accept",
              "htv:fieldValue": "text/event-stream"
            },
            {
              "htv:fieldName": "Connection",
              "htv:fieldValue": "keep-alive"
            }
          ],
          "response": {
            "htv:statusCodeNumber": 200,
            "contentType": "text/event-stream"
          }
        },
        {
          "op": "unbsubscribeevent",
          "href": "./events/overheated",
          "subprotocol": "sse"
        }
      ]
    }
  },
  "forms": [
    {
      "op": "readallproperties",
      "href": "./properties",
      "htv:methodName": "GET",
      "htv:headers": [{
        "htv:fieldName": "Accept",
        "htv:fieldValue": "application/json"
      }],
      "response": {
        "htv:statusCodeNumber": 200,
        "contentType": "application/json"
      }
    },
    {
      "op": "writemultipleproperties",
      "href": "./properties",
      "htv:methodName": "PUT",
      "contentType": "application/json",
      "response": {
        "htv:statusCodeNumber": 204
      }
    },
    {
      "op": "queryallactions",
      "href": "./actions",
      "htv:methodName": "GET",
      "htv:headers": [{
        "htv:fieldName": "Accept",
        "htv:fieldValue": "application/json"
      }],
      "response": {
        "contentType": "application/json",
        "schema": "actionStatusList"
      }
    },
    {
      "op": "subscribeallevents",
      "href": "./events",
      "subprotocol": "sse",
      "htv:methodName": "GET",
      "htv:headers": [
        {
          "htv:fieldName": "Accept",
          "htv:fieldValue": "text/event-stream"
        },
        {
          "htv:fieldName": "Connection",
          "htv:fieldValue": "keep-alive"
        }
      ],
      "response": {
        "htv:statusCodeNumber": 200,
        "contentType": "text/event-stream"
      }
    },
    {
      "op": "unsubscribeallevents",
      "href": "./events",
      "subprotocol": "sse"
    }
  ],
  "schemaDefinitions": {
    "actionStatus": {
      "type": "object",
      "properties": {
        "status": {
          "type": "string",
          "enum": [
            "pending",
            "running",
            "completed",
            "failed"
          ]
        },
        "error": {
          "type": "object"
        },
        "href": {
          "type": "string",
          "format": "uri",
          "const": "./actions/fade/{id}"
        }
      },
      "required": [
        "status"
      ]
    },
    "actionStatusList": {
      "type": "object",
      "additionalProperties": {
        "type": "array",
        "items": {
          "type": "object",
          "properties": {
            "status": {
              "type": "string",
              "enum": [
                "pending",
                "running",
                "completed",
                "failed"
              ]
            },
            "error": {
              "type": "object"
            },
            "href": {
              "type": "string",
              "format": "uri",
              "const": "./actions/fade/{id}"
            }
          },
          "required": [
            "status"
          ]
        }
      }
    }  
  }
}
```

danielpeintner:
I think people are slowly coming back from vacation (some maybe next week) and I would like to hear their opinions also (@sebastiankb @egekorkan @relu91 others).
Please don't take me wrong, I just would like to have a solution that does *not* break existing implementations...

benfrancis:
@danielpeintner wrote:
> I just would like to have a solution that does _not_ break existing implementations...

I agree that ideally it should not break existing implementations of WoT Thing Description 1.1 which don't implement the profile, even if support for all operations is not possible (because some parts of the protocol binding simply can't be described declaratively in a Thing Description).

Note that the 1.1 specification says:

_"A synchronous action means that the response of action contains all the information about the result of the action and no further querying about the status of the action is needed. Lack of this keyword means that no claim on the synchronicity of the action can be made."_

My understanding of that is that if a WoT Thing Description 1.1 consumer implements this correctly, then it shouldn't be expecting the response to the `invokeaction` request to contain the action output if `synchronous` is set to `false` (or is undefined).

It may reasonably expect the response of a synchronous action to match the output data schema, and this PR would fix that.

This isn't backwards compatible with WoT Thing Description 1.0 Consumers, but WoT Profile already says that the TD 1.1 context must be present. We could add that the WoT Thing Description 1.0 context must not be present if it helps make it more explicit that WoT TD 1.0 Consumers shouldn't try to parse them.

My expectation with the proposed text is that WoT Thing Description 1.1 consumers which don't implement the profile:
1. Should still be able to invoke both synchronous and asynchronous actions
2. Should still be able to understand the output of a synchronous action
3. Would not be able to query the status of an asynchronous action because it is not possible to describe the action queue without a profile
4. None of the action operations should generate an exception

Note that in general I don't think it's safe to assume that a WoT Thing Description 1.1 Consumer will be able to consume Things which implement a profile the Consumer doesn't support. This is because the profiling mechanism just says:

_"In order to conform with a profile, a [Web Thing](https://www.w3.org/TR/wot-architecture/#dfn-thing) MUST conform with all the normative statements in the profile's specification."_

There are no restrictions on what the normative statements in a profile can say, and therefore there's no guarantee that a Consumer which doesn't implement them will be compatible.

The best we can do is to make sure that all the profiles we define at the W3C are as compatible as possible with non-conformant Consumers. But compatibility should not be assumed, and there's nothing to stop anyone from defining their own profile which breaks compatibility altogether. Profiles are designed to promote interoperability _within_ profiles, not outside of them.

benfrancis:
@danielpeintner wrote:
> The main issue (for me) is that `output` in a profile TD may conflict with the core TD spec for async actions. That's at least how I see it.

By my understanding of the TD specification (as described above), that shouldn't be the case because a Consumer should not expect the response to an asynchronous action invocation to contain the output of the action (since it may not have completed yet). Did you interpret the specification differently?

> for async actions: What if we use a new term like actionStatusOutput that specifies the output (output remains undefined in this case).

Do you mean by extending the TD ontology? So far we have managed to avoid that in WoT Profile.

I would approach it the other way around and continue to use `output` to mean the output of the real world action, and separately define the payload of the HTTP response (where it is different) in `additionalResponses` as part of the protocol binding. See https://github.com/w3c/wot-thing-description/issues/1665

But again, that's only necessarily if you want to describe the protocol binding declaratively in the Thing Description under the `actions` member in _addition_ to the specification referenced in the `profile` member.

If you only rely on the profile then a non-conformant Consumer will still be able to invoke an asynchronous action but should not try to read its output unless there is also a `queryaction` operation defined on the action affordance, with a Form describing how to use it.

In other words, I don't think there's an incompatibility.

I do think that the TD specification could be clearer (in the informative table in section 5.3.4.2.1) that the `output` data schema only maps onto the response of the `invokeaction` operation when `synchronous` is set to `true`. I have suggested that in https://github.com/w3c/wot-thing-description/issues/1665.

Note: This is all only a problem in request/response style protocols like HTTP and CoAP where there is always assumed to be a response to every operation. For something like WebSockets, or pub/sub style protocols it probably isn't even an issue.

egekorkan:
A lot to discuss here. Some general thoughts:
1. `output` is not defined properly in the TD spec. It can be the response to the request or the actual result of the action. The term output is vague, I am not sure if this is by design. In the table I have created (https://w3c.github.io/wot-thing-description/#sec-op-data-schema-mapping) I am sort of implying that the output is the message sent by the Thing to the Consumer but even this does not mean that the message is the response to the initial invokeaction request.
2. As Ben said at "Note: This is all only a problem in request/response style protocols like HTTP and CoAP where there is always assumed to be a response to every request. For something like WebSockets, or pub/sub style protocols it probably isn't even an issue." all this discussion takes a different direction in a protocol like MQTT which does not even have a way to return a message to the Consumer about the result of the action. Only in MQTT5 there is a way but we did not explore it properly yet. In node-wot, an error or warning is thrown if you want to return a response to invokeaction.
3. Non Profile Consumers being able to understand profile: I think that we are too late for this. If we want this, this has to be a requirement for all profiles, present and future. This would imply that all profile TDs have to be subset of the TD spec and no behavior can be mandated in a profile since a TD cannot describe this kind of behavior. I think that implementations like node-wot need to invent a new mechanism similar to how different protocols are handled for profiles, meaning a new "binding" per profile (yay interoperability...). Each profile implementation will use a new codebase. We are lucky that this profile is rather simple but there is nothing stopping a profile to describe that in order to invoke an action, the consumer needs to send 3 GET requests and then 4 POST requests or mixing different protocols in there. @danielpeintner the "bindings" for fujitsu and oracle inside node-wot were doing some custom logic even though that it was just http and websocket in the background. 

Some solutions:
1. Giving up trying to make non-Profile Consumers (like node-wot) consume and interact profile-compliant Things.
2. Trying to adapt to non-Profile Consumers:
  1. Declaratively writing the response of the invokeaction inside the output no matter sync or async. A profile compliant would need to parse schema information inside forms to understand the type of the actual result of the action. A non-profile Consumer will not be able to query or cancel but they cannot do that now anyways.
 
I think @danielpeintner wants 2.1 but it makes for weird TDs. Why does matter that a non-profile Consumer can understand the invokeaction response for an async action, it will not be able to do anything with it anyways.

danielpeintner:
I fully agree with the statement that 

>  all profile TDs have to be subset of the TD spec

which is not the case!

Anyhow, maybe it is really too late to achieve that.

One more note. The reason *why* node-wot looks into the result of an action is to make sure that bindings report the *expected* value (type). We do that for properties, action and events by validating JSON values with AJV.
Anyhow, if people think this should **not** be done we might need to rethink this approach..

In general, I do not want to push-back the profile document. I just want to point you to some problems I see. In the end it boils down to the fact that the profile spec is not a subset of the TD spec.

benfrancis:
@danielpeintner wrote:
>> all profile TDs have to be subset of the TD spec
> which is not the case!

To be clear, the Thing Descriptions are valid TD 1.1 Thing Descriptions, there is no superset of the TD Information Model in WoT Profile (that's something I argued against).

It's true that the concrete protocol bindings defined in WoT Profile are a superset of what it's currently possible to describe using protocol binding templates, due to limitations of the Thing Description specification. However, the example TDs in the WoT Profile specification are designed such that even Consumers which don't implement the profile will still be able to consume a subset of functionality, and ideally shouldn't generate any exceptions. Issue #259 identified one area where this could be improved.

With the change proposed in this PR, the response to a synchronous action invocation request will be what a generic Consumer would expect, i.e. it will map onto the output data schema of the ActionAffordance.

> The reason why node-wot looks into the result of an action is to make sure that bindings report the expected value (type). We do that for properties, action and events by validating JSON values with AJV. Anyhow, if people think this should not be done we might need to rethink this approach..

Does node-wot behave any differently for ActionAffordances where `synchronous` is set to `false`? If it assumes that the output of the action is contained within the response to the action invocation HTTP request, then that is surely incorrect. By definition, that will not be the case for an asynchronous action.

I'm not saying that this is the fault of node-wot developers, I think it's an ambiguity in the Thing Description specification about what "output" means, which I have proposed a solution to in https://github.com/w3c/wot-thing-description/issues/1665. As @egekorkan said, "`output` is not defined properly in the TD spec".

For the purposes of this PR, I guess the question is whether the change proposed here is an improvement? If it is, then let's get it landed and I can work on adding different or additional examples which show action outputs.

As for whether we can live with the compromises made in WoT Profile 1.0 as the specification progresses to a Candidate Recommendation, that's another matter which needs further discussion.

benfrancis:
@egekorkan wrote:
> Some solutions:
> 1. Giving up trying to make non-Profile Consumers (like node-wot) consume and interact profile-compliant Things.

I don't think we need to give up on that altogether, but Consumers will inevitably only be able to use a subset of the Thing's functionality if they don't implement the profile because a Thing Description can not fully describe an asynchronous action queue.

> 2. Trying to adapt to non-Profile Consumers:
> 3. Declaratively writing the response of the invokeaction inside the output no matter sync or async. A profile compliant would need to parse schema information inside forms to understand the type of the actual result of the action. A non-profile Consumer will not be able to query or cancel but they cannot do that now anyways.

In my view Consumer implementations should not expect the response to an asynchronous action invocation HTTP request to map onto the output data schema. For asynchronous actions the output data schema should be assumed to be used by a `queryaction` operation, not an `invokeaction` operation.

For Things which have asynchronous actions but no action queue (i.e. only one invocation is allowed at a time), it should actually be possible to describe the protocol binding using a `queryaction` operation in a Thing Description.

In the case of an asynchronous action queue like the one defined in the HTTP Baseline Profile, it's not currently possible to describe the protocol binding using a `queryaction` operation in a Thing Description because there's no way to unambiguously describe the dynamically created resources. Therefore, the Thing Description should not include a `queryaction` operation and the assumptions are instead baked into the profile.

This means that a generic Consumer will be able to invoke an asynchronous action, but not query its status. I personally think this is fine, because one of the big benefits of profiles is that they can provide a short hand for greenfield Things which all use the same protocol binding. We should embrace this as a benefit and not limit profiles by the limitations of Thing Descriptions (which will never be able to describe all possible protocol bindings).

PR #271, opened 2022-09-02
WIP: common constraints - identifiers

mlagally:


index.html
```diff
@@ -1854,6 +1854,53 @@ <h2>Default Language</h2>
           (e.g., en, de-AT, gsw-CH, zh-Hans, zh-Hant-HK, sl-nedis).</span>
         </span>
     </section>
+    <section id="sec-identifiers">
+      <h2>Identifiers</h2>
+      <p>
+        Web resources are identified by uniform resource identifiers (URIs).
+        The Thing Description defines the field <code>id</code> for that purpose, which is in the format "anyURI".
```
benfrancis, line 1861: Reference needed for "anyURI" (or the `id` member in TD which in turn has a reference for "anyURI")

index.html
```diff
@@ -1854,6 +1854,53 @@ <h2>Default Language</h2>
...
+      <p>
+        A common and well defined format for these identifiers enables easy implementation 
+        and simplifies handling across manufacturers through the use of libraries. 
+        It also ensures that there are no name clashes of devices from different manufacturers.
+        The [[UUID]] format has been developed to satisfy these requirements.
+      </p>
+        <p>
+          <span class="rfc2119-assertion" id="common-constraints-identifiers-uuid">
+            It is RECOMMENDED to use Universally Unique IDentifier (UUID) [[RFC4122]] as common format for identifers.
+          </span>
```
benfrancis, line 1877: I'm pleased to see that this is only a recommendation, however:
1. A UUID is not a URI. Do you perhaps mean a URN?
2. Can you explain why a URN is a better globally unique identifier than an HTTP URL with a fully qualified domain name and therefore a URN should be recommended over that?

In WebThings we still use the URL of the Thing Description as its ID, which has many benefits.

index.html
```diff
@@ -1854,6 +1854,53 @@ <h2>Default Language</h2>
...
+        </p>
+          <span class="rfc2119-assertion" id="common-constraints-identifiers-random">
+          It is RECOMMENDED to use [[UUIDv4]] as common format for unique identifers, where random identifiers are required.
+          </span>
+      </p>
+      </section>
+      <section id="sec-identifiers-mutability">
+        <h3>Unique Identifiers</h3>
```
benfrancis, line 1885: Should this title be "Mutable Identifiers" or "Immutable Identifiers"?

index.html
```diff
@@ -1854,6 +1854,53 @@ <h2>Default Language</h2>
...
+      <p><span class="rfc2119-assertion" id="common-constraints-identifiers-mutable">
+        In certain use cases where privacy is a factor or legal obligations mandate changing identifiers, 
+        the use of a <em>mutable UUID</em> is RECOMMENDED.</span>
+      </p>
+
+      <p><span class="rfc2119-assertion" id="common-constraints-identifiers-immutable-ids">
+        If possible and allowed by law, required immutable IDs SHOULD be made available through a
+        property rather than used as the id for the TD itself.
+      </span></p>
+      </section>
```
benfrancis, line 1903: This is already covered in section [11.2 Immutable Identifiers](https://w3c.github.io/wot-thing-description/#sec-privacy-consideration-id) of TD 1.1. I don't think this text really adds anything to that text.

benfrancis (commented):
I don't particularly object to this section, but the mutable/immutable stuff is already covered in the TD spec and I'm not convinced that a URN is necessarily better than other URIs, like an HTTP URL with a fully qualified domain name for example.

index.html
mlagally: The URL is not stable in scenarios where things may transition between different networks. Think of roaming scenarios in fleet management, IP reassignment cases etc.

index.html
mlagally: This goes beyond what the TD describes in section 11.2.

The proposed text talks specifically about UUIDs, i.e. a format this is well-defined, whereas the TD does not have any format description for identifiers.

index.html
mlagally: Both are valid titles, I prefer "Mutable identifiers"

index.html
mlagally: reference was added.

benfrancis:
- A UUID is not a valid URI
- The HTML in this PR doesn't render correctly
- The proposed text is very verbose and a lot of it is redundant
- I think best practice recommendations like this really belong in the Thing Description specification since the apply to all Thing Descriptions, not just those which conform with a profile

I don't have a problem with recommending URNs as IDs, as long as any URI is allowed.

If you really feel that it's important for this recommendation to be made in the Profile specification, I suggest reducing this section down to a single assertion which says something along the lines of:

"It is RECOMMENDED that a Web Thing's Thing Description uses a URN [[URN]] as the value of its `id` member, so that it has a globally unique identifier."

egekorkan:
I agree with @benfrancis but the most important thing here is to remove the last assertion since the TD spec says:

> Ideally, any required immutable identifiers SHOULD only be made available via affordances, such as a property, whose value can only be obtained after appropriate authentication and authorization, and managed separately from the TD identifier.

benfrancis (changes requested):
See my comments above.

PR #267, opened 2022-08-24
Replace Information Model section with Best Practices section - closes #10

benfrancis:
This is a follow-up to #260 which attempts to resolve the long-running issue #10 regarding the extent to which profiles constrain the Thing Description Information model, and is a replacement for PR #125.

The work so far in #260 includes many of the constraints from the original Core Data Model section in six sub-sections of a new Common Constraints section:
1. Security
2. Discovery
3. Links
4. Errors
5. Semantic Annotations
6. Default Language

My proposal in this PR is to replace the content of the remaining Information Model section (what's left from the original "Core Data Model" section) with a Best Practices section which includes some of the key remaining constraints as recommendations.

The content of the current Information Model section is still very contentious and we are running out of time to agree on which constraints to keep. This proposal is a compromise which retains some of the key constraints from that section as recommendations which I think we can get quick consensus on.

I have explained the rationale behind the removal of other more rigid constraints in great detail in other issues and PRs, but I'm happy to answer any questions about constraints that others feel should be kept.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/267.html" title="Last updated on Aug 31, 2022, 3:22 PM UTC (df94505)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/267/f7348cf...benfrancis:df94505.html" title="Last updated on Aug 31, 2022, 3:22 PM UTC (df94505)">Diff</a>

PR #272, opened 2022-09-05
Draft: Profile 1.0 Spec

mlagally:
This branch contains a proposal of how the profile 1.0 release can be structured. 

Scope has been reduced to only normatively define the HTTP baseline profile. It makes the two event profiles optional.
Language has been cleaned up, the introduction section has been improved, the two event profiles were made informative and corresponding language has been provided.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/272.html" title="Last updated on Sep 5, 2022, 11:34 AM UTC (1abd758)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/272/f7348cf...1abd758.html" title="Last updated on Sep 5, 2022, 11:34 AM UTC (1abd758)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -678,21 +678,18 @@
       The following sections are applicable for all of the HTTP profiles defined by this document.
     </p>
 
-    <!-- Identifiers -->
-
-
+    <!-- Identifiers TODO -->
 
     <!-- Units -->
 
     <section id="common-constraints-units">
       <h2 id="units">Date format</h2>
-      <span class="rfc2119-assertion" id="common-constraints-units">
-        It is highly RECOMMENDED to always specify a
-        <code>unit</code>, if a value has a metric.</span>
-      Authors of <em>Thing Descriptions</em> should be aware that units
-      that are common in their geographic region are
-      not globally applicable and may lead to
-      misinterpretation with drastic consequences.
+      <p>Authors of <em>Thing Descriptions</em> should be aware that units
+      that are common in their geographic region are not globally applicable 
+      and may lead to misinterpretation with drastic consequences.
+      </p>
+      <p><span class="rfc2119-assertion" id="common-constraints-units">
+        It is highly RECOMMENDED to specify a <code>unit</code>, if a value has a metric.</span>
       </p>
     </section>
     <!--  Date Format -->
--- a/index.html
+++ b/index.html
@@ -678,23 +678,35 @@
       The following sections are applicable for all of the HTTP profiles defined by this document.
     </p>
 
-  <!--  TODO: Identifiers -->
+    <!-- Identifiers -->
 
-    <!--  TODO: Units -->
 
+
+    <!-- Units -->
+
+    <section id="common-constraints-units">
+      <h2 id="units">Date format</h2>
+      <span class="rfc2119-assertion" id="common-constraints-units">
+        It is highly RECOMMENDED to always specify a
+        <code>unit</code>, if a value has a metric.</span>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -171,8 +177,9 @@
       other features of the Thing Description that go beyond the constraints
       of the HTTP Baseline Profile, however the interoperability guarantees of the HTTP Baseline Profile
       hold only for the <a>HTTP Baseline Profile</a> subset.
-
-
+    </p>
+    <p>Future versions of this document may contain other profiles,
+      e.g. a profile for digital twins and a profile for resource constrained devices.</p>
```
benfrancis, line 182: I would prefer not to list digital twins as an example because I'm not yet convinced that's a good example use case for a profile. Better to just use the constrained devices example which is less contentious.

index.html
```diff
@@ -759,796 +678,43 @@ <h1>Common Constraints</h1>
...
+      </p>
+      <p>
+        As described by this section the following constraints must be observed:
+        <span class="rfc2119-assertion" id="common-constraints-date-format-2">
+          All dateTime values SHOULD use UTC as the time zone and use the 'Z' identifier.</span>
+      </p>
+      <p>
+        <span class="rfc2119-assertion" id="common-constraints-date-format-3">
+          A time value of 24:00 is not permitted; the value 00:00 MUST be used instead.</span>
+      </p>
```
benfrancis, line 717: In specref, [[xmlschema-2]] points to https://www.w3.org/TR/xmlschema-2/ which doesn't appear to have a section 3.2.7.

That spec does say _"'24' is permitted if the minutes and seconds represented are zero"_, which is not what you want.

In https://github.com/w3c/wot-profile/pull/267 I proposed using the reference [[RFC3339]] (https://www.rfc-editor.org/rfc/rfc3339) instead, since that already says:

_"Although ISO 8601 permits the hour to be "24", this profile of ISO 8601 only allows values between "00" and "23" for the hour in order to reduce confusion."_

RFC 3339 also only supports UTC and only numerical offsets (not those that don't have an integral number of minutes). I believe this resolves the ambiguity of timezones you are concerned about, without being quite so prescriptive.

In short, if you just reference RFC3339, the rest of the qualifying text is not needed.

index.html
```diff
@@ -759,796 +678,43 @@ <h1>Common Constraints</h1>
       The following sections are applicable for all of the HTTP profiles defined by this document.
     </p>
 
-    <section id="http-profiles-information-model">
-      <h2>Information Model</h2>
```
benfrancis, line 763: 🎉

benfrancis (commented):
There are two separate large changes in this PR...

The first appears to be an alternative to https://github.com/w3c/wot-profile/pull/267 and replaces the Information Model section with separate smaller sections which are less contentious. I strongly support this change, and actually prefer it to my own PR.

The second is to make the HTTP SEE Profile and HTTP Webhook Profile informative. I think that requires further discussion.

index.html
mlagally: Will change to "profiles for other protocols and profiles for resource constrained devices."

mlagally: Profile call on Wed 7h: remove lines 180-182

mlagally:
Agreed to merge in Profile call on Sept 7th.

index.html
mlagally: This is text we previously had agreed on. If it needs to be improved, let's treat it in a separate issue.
I filed a new issue for a follow up.

Issue #229, opened 2022-06-17
remove http-webhook-profile-protocol-binding-general-2 assertion


egekorkan:
There is an assertion with id `http-webhook-profile-protocol-binding-general-2` and text `The HTTP Webhook profile MAY be used as an alternative event mechanism to the [HTTP SEE Profile](https://w3c.github.io/wot-profile/#sec-http-sse-profile).`

I think this assertion is not necessary and it cannot be tested in any way. I think it should be removed

mlagally:
@egekorkan Please explain why you think it is not necessary.
We define 2 event mechanisms, Each of them MAY be used or they MAY be used together.

egekorkan:
Any assertion that cannot be tested (automatic or manual), thus asserted, cannot be in a spec and it's actually not an assertion. I'm not saying that this doesn't make sense, it should simply not be an RFC assertion.

egekorkan:
Any assertion that cannot be tested (automatic or manual), thus asserted, cannot be in a spec and it's actually not an assertion. I'm not saying that this doesn't make sense, it should simply not be an RFC assertion.

PR #277, opened 2022-09-07
remove http-webhook-profile-protocol-binding-general-2 assertion - closes issue #229

mlagally:
resolves issue #229

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/277.html" title="Last updated on Sep 7, 2022, 1:25 PM UTC (889e12a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/277/8b8e1aa...889e12a.html" title="Last updated on Sep 7, 2022, 1:25 PM UTC (889e12a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2813,8 +2813,8 @@
         provide operations to read and write properties and invoke, query and
         cancel actions.</span>
     </p>
-    <p><span class="rfc2119-assertion" id="http-webhook-profile-protocol-binding-general-2">
-        The <em>HTTP Webhook profile</em> MAY be used as an alternative event mechanism
+    <p>
+        The <em>HTTP Webhook profile</em> may be used as an alternative event mechanism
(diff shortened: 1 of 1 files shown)


egekorkan (changes requested):
No need for may, I would simply say that:
HTTP Webhook Profile is an alternative event mechanism to SSE profile

mlagally:
> No need for may, I would simply say that:
> HTTP Webhook Profile is an alternative event mechanism to SSE profile

This does not change the meaning at all. Why do you request this change?

egekorkan:
> Why do you suggest this change?

I do acknowledge that if it is not marked in a span, it is not an official assertion to be tested. However, that language is invented for a reason, and using it non-capitalized is simply escaping the testing burden. Sometimes we cannot avoid the language but here we can so why not :)

Also, a future editor or reviewer can stumble on this again and have the same problem as I have.

PR #287, opened 2022-09-21
Consolidate date-time references to RFC 3339 - closes #276

### LINKED ISSUE #276: Clarifications on DateTime format
Carry-over from a comment by @benfrancis :

In specref, [[xmlschema-2]] points to https://www.w3.org/TR/xmlschema-2/ which doesn't appear to have a section 3.2.7.

That spec does say "'24' is permitted if the minutes and seconds represented are zero", which is not what you want.

In https://github.com/w3c/wot-profile/pull/267 I proposed using the reference [[RFC3339]] (https://www.rfc-editor.org/rfc/rfc3339) instead, since that already says:

"Although ISO 8601 permits the hour to be "24", this profile of ISO 8601 only allows values between "00" and "23" for the hour in order to reduce confusion."

RFC 3339 also only supports UTC and only numerical offsets (not those that don't have an integral number of minutes). I believe this resolves the ambiguity of timezones you are concerned about, without being quite so prescriptive.

In short, if you just reference RFC3339, the rest of the qualifying text is not needed.

benfrancis:
This PR addresses issue #276. It fixes inconsistent and broken references and changes all date-time constraints within the specification to refer to the Date Time section, which itself references RFC 3339.

I have changed the clarifying assertions in the Date Time section into a Note since they only repeat what it says in RFC 3339.

Apart from making the date format consistent throughout the specification, the only significant normative change is that timezones _are_ allowed in timestamps, but only by way of unambiguous numerical UTC offsets as allowed by RFC 3339. This still helps remove ambiguity, but fixes the awkward situation where some RFC 3339 timestamps are valid and others are not, which should help simplify implementation.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/287.html" title="Last updated on Sep 21, 2022, 10:39 AM UTC (8a2e8fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/287/8b8e1aa...benfrancis:8a2e8fe.html" title="Last updated on Sep 21, 2022, 10:39 AM UTC (8a2e8fe)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -700,20 +700,20 @@
     <!--  Date Format -->
 
     <section id="common-constraints-date-format">
-      <h2 id="date-format">Date format</h2>
+      <h2 id="date-format">Date Format</h2>
       <p>
         <span class="rfc2119-assertion" id="common-constraints-date-format-1">
-          All date and time values MUST use the canonical dateTime representation format
-          defined in section 3.2.7 of [xmlschema-2].</span>
+          All date and time values MUST use the <code>date-time</code> format 
+          defined in [[RFC3339]].</span>
       </p>
-      <p>
-        As described by this section the following constraints must be observed:
-        <span class="rfc2119-assertion" id="common-constraints-date-format-2">
-          All dateTime values SHOULD use UTC as the time zone and use the 'Z' identifier.</span>
-      </p>
-      <p>
-        <span class="rfc2119-assertion" id="common-constraints-date-format-3">
-          A time value of 24:00 is not permitted; the value 00:00 MUST be used instead.</span>
+      <pre class="example">
+        <code>2022-09-21T23:20:50.52Z</code>
+      </pre>
+      <p class="note">
+        In order to reduce ambiguity, RFC 3339 only permits an hour with a value
+        between 00 and 23 (not 24), and time zones expressed as a numerical
+        offset relative to UTC. The suffix "Z" when applied to a time denotes a
+        UTC offset of 00:00.
       </p>
     </section>
 
@@ -1616,8 +1616,8 @@
                   <td><code>timeRequested</code></td>
                   <td>
                     A timestamp indicating the time at which the Thing received
-                    the request to execute the action, in ISO format [[ISO8601-1]]
-                    (see <a href="#date-format">Date format</a> for additional
+                    the request to execute the action. (See 
+                    <a href="#date-format">Date Format</a> for date format 
                     constraints).
                   </td>
                   <td>optional</td>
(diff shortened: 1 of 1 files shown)


mlagally (changes requested):
The proposed text replaces the reference to XMLSchema, which refers to ISO8601.

The proposed alternative date format reference is to RFC 3339, which is slightly different,
see for a discussion: https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats#:~:text=ISO%208601%20allows%20for%20things,doesn't%20allow%20for%20that.

This has some significant drawbacks, e.g. widely used runtimes use the ISO-8601 format:

https://docs.oracle.com/javase/10/docs/api/java/time/package-summary.html

I think a direct reference to ISO-8601 would be  proper fix - The clarifications on the hour value of 24 and constraints to UTC are useful and should be preserved.

benfrancis:
Currently the specification has a mix of references to [xmlschema-2](https://www.w3.org/TR/xmlschema-2/), [ISO8601-1](https://www.iso.org/standard/70907.html) and [RFC3339](https://www.rfc-editor.org/rfc/rfc3339).

This PR is intended to consolidate them around a single format for consistency.

As I understand it RFC 3339 is a "profile" of the ISO 8601 standard which adds additional constraints to remove ambiguities (such as not allowing 24 as an hour), which seems appropriate here.

Note that the Thing Description specification [says](https://w3c.github.io/wot-thing-description/#td-basic-types-mapping) that:

> Values that are of type dateTime MUST be serialized as JSON strings following the "date-time" format specified by [[RFC3339](https://w3c.github.io/wot-thing-description/#bib-rfc3339)]

Given that some timestamps which are valid ISO 8601 are not valid under RFC 3339, referencing ISO 8601 is not really enough.

This is a notoriously complicated area and it's also true that there are some timestamps which would be valid under RFC 3339 but not valid under ISO 8601. There's a useful comparison [here](https://github.com/ijmacd/rfc3339-iso8601) and visualisation [here](https://ijmacd.github.io/rfc3339-iso8601/).  As I understand it most programming languages only support a simplified subset of ISO 8601. For example, ECMA262 specifies a [simplified version](https://tc39.es/ecma262/#sec-date-time-string-format) of ISO 8601 which can be parsed by JavaScript, so allowing any ISO 8601 timestamp also unfortunately does not guarantee interoperability.

My suggestion is just to refer to an existing standard rather than invent our own, and being consistent with the Thing Description specification seems like a good idea.

mlagally (approved):

mlagally:
Arch call on Nov 10th:
Agree to merge

PR #289, opened 2022-09-21
Remove Links section - closes #255

### LINKED ISSUE #255: Review of section 6.1.2 Links
I'm sorry I'm only just getting to review this after PR #240 has already landed.

@mlagally Did you take into account the previous discussion on this topic in #114?

Review comments below.

### 6.1.2 Links

> The following keywords are defined for the HTTP profiles and MUST be supported by all consumers.

In principle I like the idea of defining a finite set of link relation types which Consumers MUST support and Producers MAY use, but I'm not sure how it can work in practice (see below).

What does "keywords" mean here? Does it mean Consumers must support all of the included members of the link object, and all of the listed link relation types? I'm going to assume that's the case.

> type

WebThings currently uses links to link to WebSocket endpoints, which don't have a `type` member, just a `href` and a `rel` member, because the type member wouldn't make sense.

We are [discussing](https://github.com/WebThingsIO/gateway/issues/2882) changing these links to forms, but there are still unresolved issues in the Thing Description specification which may prevent that.

I suggest it would be fine for `type` to remain optional as in the Thing Description specification. It being mandatory doesn't really gain anything anyway as far as I can tell (see below).

> hreflang

I see there's already an issue filed to make this optional, which I agree with.

#### 6.1.2.1 Link Relation Types

How would all of the assertions in this section be tested?

> icon

What does it mean have mandatory support for icons? What if a Consumer doesn't have a screen on which to render an icon? The specification limits icons to PNG and JPEG. What about WEBP, GIF and SVG, which are also widely supported on the web? For WebThings we prefer SVG icons for Things so that they are scaleable, so would want SVG support and the "any" value for sizes.

> type

What does a Consumer actually need to do with this kind of link in order to be conformant? My understanding is that Thing Models are only used internally by applications for generating Thing Descriptions and are not intended to be used by Consumers in any way.

> service-doc

Again, what if a Consumer can't render, Markdown, HTML or PDF files?

> collection

What would the mandatory type value be for this kind of link relation? Does it have to be a Thing Description, a Directory Description, or can it be any kind of resource? (See: https://github.com/w3c/wot-thing-description/issues/1567).

What must a Consumer do with this kind of link relation in order to be compliant? Present a clickable link to the collection to the user? What happens when the user clicks that link? What if there is no user interface?

> item

Does the target type have to be a Thing Description? What must a Consumer do with this kind of link relation in order to be compliant?

#### 6.1.2.2 Media Types for Link Targets

> The following media types from [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml) are defined for the link targets of compliant TDs. They MUST be supported by all consumers.

What must a Consumer do with these links in order to comply with this assertion? This seems like quite an arbitrary list and I have many questions.

Does a Consumer have to be able to render text, HTML and PDF documents? Do all Consumers have to have a web rendering engine in order to render HTML? What if they don't have a screen?

Do all Consumers need to implement a markdown renderer? Which version of markdown? There are many.

What must a Consumer do with machine-readable JSON files, Thing Descriptions and octet streams?

Must a Consumer be able to render JPEGs and PNGs? What if it doesn't have a screen?

---

In WebThings, aside from the WebSocket links which we would like to replace with Forms, our main use case for Links is providing a link to an HTML user interface for a Thing, which is the example I provided in #114. E.g.:

```
{
  "href": "https://user.webthings.io/things/mything1",
  "rel": "alternate",
  "type": "text/html"
}
```

I would like it if we could make support for this part of the HTTP Profiles, but again what would it mean that it is mandatory for all Consumers to support this? What happens if a Consumer can't render an HTML page?

In conclusion, whilst in principle I like the idea of a finite set of link relation types that Consumers MUST support and Producers MAY use, in practice I'm not sure it would actually work. I'm struggling to imagine how these assertions can be implemented, or tested.

benfrancis:
This PR proposes the removal of the Links section from Common Constraints.

This is the only solution I can think of to resolve the issues raised in #255, because in its current state the assertions are not testable or implementable by all Consumers.

I'd be open to improvements in this section if it can be made implementable, as I agree with the desire to standardise certain link relation types (e.g. for HTML user interfaces and icons), I just can't think of any way to do that which adds any value beyond the existing recommendations in the Thing Description specification.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/289.html" title="Last updated on Sep 28, 2022, 10:54 AM UTC (19a037a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/289/3d4a8cb...benfrancis:19a037a.html" title="Last updated on Sep 28, 2022, 10:54 AM UTC (19a037a)">Diff</a>


mlagally:
@benfrancis
I consider it very bad practice to just create a PR that removes an entire section.
Please consider creating a constructive proposal to improve the current text.

egekorkan (approved):
I find the PR good and motivation of it reasonable. Not sure if there is any implementations of it anyways (since these are about Consumers and we have none at the moment?)

mlagally (changes requested):
There are relations that are useful for handling collections of things.
These enable constructing things that consist of things.

We should clarify what we mean with "a consumer must handle" these types.
The intention of the current text is not to enforce a specific processing (e.g. a printer thing certainly does different actions on a PDF document than a screen or a device without screen).
The intention of this section is to define a clear and unambiguous **signaling** of these types, to **enable** appropriate handling by consumers.

Please take this into account in this PR.

benfrancis:
@mlagally wrote:
> There are relations that are useful for handling collections of things.
> These enable constructing things that consist of things.

Yes, these link relation types are described in the Thing Description specification. The current assertions in WoT Profile make it mandatory for Consumers to "support" those link relation types, but the assertions don't explain what "supporting" them involves, and therefore can't be tested.

> We should clarify what we mean with "a consumer must handle" these types.

Agreed.

> The intention of this section is to define a clear and unambiguous signaling of these types, to enable appropriate handling by consumers.

What does "signaling" involve? How can I tell whether my consumer implementation supports this or not? What are the requirements?

I still the think the list of link relation types is arbitrary and impossible to implement on all consumers.

mlagally:
@benfrancis , @egekorkan 
Please review #315 which provides improved language.
This may need further clarifications, but I hope it explains the added value of defining the link types a consumer can expect from a compliant thing.

benfrancis:
I still think the Link section has issues, with "MAY" assertions that don't really have an effect and arbitrary lists of values (and one missing value), but the new assertions at the end of the section at least provide something for Consumers to implement. I therefore don't think that removing the section entirely is necessary so I'm going to close this PR.

PR #293, opened 2022-09-21
Allow security metadata on Forms - closes #292

### LINKED ISSUE #292: Re-consider allowing security metadata on Forms
Currently there are two assertions in section [6.3 Security](https://w3c.github.io/wot-profile/#common-constraints-security) which say:
- Security schemes MUST be applied using the top level security member of a [Thing](https://w3c.github.io/wot-thing-description/#thing).
- Security schemes MUST NOT be applied to individual Forms.

I understand why this is desirable in order to reduce implementation complexity of Consumers, but I'm concerned about potential implications of that constraint.

Implementation experience with WebThings has shown me that sometimes different authentication schemes need to be used for different protocols or sub-protocols. For example, an implementation of the Basic HTTP Profile may use the OAuth2 or Bearer security scheme with a JSON Web Token in an `Authorization` header (`in: "header"`), but an implementation of the HTTP SSE Profile may have to include that token in a query string (`in: "query"`), because web browsers do not support Authorization headers with EventSource. That is also the case for WebSockets.

In order to use multiple protocol bindings or multiple profiles in the same Thing Description (e.g. the HTTP Basic Profile and HTTP SSE Profile, or an HTTP protocol binding and WebSocket protocol binding), it may be necessary to provide multiple `securityDefinitions`, and declare on a form by form basis which security definition should be used for a given protocol binding.

If I understand correctly, if the [`security` member of Thing](https://w3c.github.io/wot-thing-description/#thing) has multiple values in an array it means that _all_ of those security schemes have to be satisfied in order to access resources. But what is needed in this case is different security definitions for different protocols or sub-protocols.

If these assertions remain, it may make it impossible to implement Things with certain combinations of profiles or protocol bindings.

I suggest removing these assertions.

benfrancis:
This PR proposes removing two assertions regarding where security metadata is allowed.

Closes #292.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/293.html" title="Last updated on Sep 21, 2022, 4:37 PM UTC (f83d835)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/293/5912319...benfrancis:f83d835.html" title="Last updated on Sep 21, 2022, 4:37 PM UTC (f83d835)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -753,17 +753,6 @@
         <p><span class="rfc2119-assertion" id="common-constraints-security-3">
             A Thing MAY implement multiple security schemes.</span>
         </p>
-        <p><span class="rfc2119-assertion" id="common-constraints-security-4">
-            Security schemes MUST be applied using the top level <code>security</code>
-            member of a
-            <a href="https://w3c.github.io/wot-thing-description/#thing">Thing</a>.</span>
-        </p>
-        <p><span class="rfc2119-assertion" id="common-constraints-security-5">
-            Security schemes
-            MUST NOT be applied to individual
-            <a href="https://w3c.github.io/wot-thing-description/#form"></a><code>
-                Form</code>s.</span>
-        </p>
     </section>
 
     <!-- Discovery -->


mlagally:
After thinking this throug, the proposed change provides a lot of additional flexibility.

mlagally (approved):

mlagally:
Arch call on Nov 3rd: Agree to merge

PR #294, opened 2022-09-21
Make security bootstrapping mandatory - closes #250

### LINKED ISSUE #250: Consider making security bootstrapping mandatory in the HTTP Baseline Profile
This is a follow-up to #121 (which adds a Discovery section to the HTTP Baseline Profile), to address the following issue:

> The WoT Discovery specification now describes another mechanism in [7.1.2 Security Bootstrapping](https://w3c.github.io/wot-discovery/#exploration-secboot) which uses HTTP 401, 302 or 303 response codes (depending on the security mechansim being used). Support for this bootstrapping mechanism is not currently mandatory for either Producers or Consumers, so as things stand only unauthenticated Thing Descriptions (consumed by Consumers which support HTTP and a Direct introduction mechanism) are guaranteed to be discoverable with the above addition to the HTTP Baseline Profile.

In order to support the discovery of authenticated Thing Descriptions (which I imagine in practice will be the majority of Thing Descriptions), we could make support for the [security bootstrapping mechanism](https://w3c.github.io/wot-discovery/#exploration-secboot) defined in the WoT Discovery specification mandatory for Consumers complying with the HTTP Baseline profile.

A bootsrapping mechanism would need to be supported for each of the mandatory security schemes (see also: #220).

benfrancis:
This PR makes [security bootstrapping](https://w3c.github.io/wot-discovery/#exploration-secboot) mandatory for:
- Consumers (for all the security schemes they implement)
- Things which require authentication in order to access their Thing Description

Closes #250.

Without these assertions, Consumers will only be out-of-the-box interoperable with Thing Descriptions which don't require authentication, which I expect will be a minority of Thing Descriptions.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/294.html" title="Last updated on Nov 3, 2022, 10:49 AM UTC (001fe9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/294/3ba8cda...benfrancis:001fe9a.html" title="Last updated on Nov 3, 2022, 10:49 AM UTC (001fe9a)">Diff</a>


mlagally (approved):
I agree with the proposed wording for now, but please note that currently all normative language in the referenced section of the discovery spec is at risk. Without sufficient implementation experience this chapter is more or less informative.

In this case the Profile spec must have a fallback plan.

mlagally:
Arch call on Nov 3rd. Agree to merge after resolving merge conflict

PR #295, opened 2022-09-26
Prepare Testfest/Plugfest - informative assertion markup - generate manual.csv

mlagally:
This PR prepares the specification draft and provides the list of normative assertions that are to be tested and verified in the Testfest/Plugfest.

Changes:

1. Modify the normative RFC2119 markup in informative sections in the document to be informative. 
2. Generate assertions.csv and manual.csv
3. Minor script modifications

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/295.html" title="Last updated on Sep 28, 2022, 9:32 AM UTC (54c21bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/295/5912319...54c21bc.html" title="Last updated on Sep 28, 2022, 9:32 AM UTC (54c21bc)">Diff</a>


egekorkan (approved):

egekorkan:
I guess it is fine but profile testing should not be a copy paste of the TD testing in my opinion. Testing some TDs does not show any OOTBI

k-toumura (approved):

testing/manual.csv
```diff
@@ -1 +1,82 @@
-"ID","Status","Comment"
+"Index","ID","Status","Comment","Assertion"
```
sebastiankb, line 1: In general, this looks good, but we should also consider a interop test by two independent implementations since this is a n important aspect of the profile.  Maybe this should be considered in the manual assertions?

benfrancis:
> Modify the normative RFC2119 markup in informative sections in the document to be informative.

I personally think it's too early to decide which parts of the specification informative as a strategy for allowing the specification to proceed to CR. I don't think there's yet sufficient evidence that the HTTP Basic Profile is any better supported than the HTTP SSE Profile for example.

But as long as _all_ of the assertions are included in test reports then I suppose we can revert these changes later if necessary.

mlagally:
Arch call on Oct. 19th.
Close without merging, this is superseeded by #296

PR #296, opened 2022-09-28
full list of assertions, including those for informative SSE and Webhook profiles

mlagally:


danielpeintner:
I think there are some conflicts popping up. This PR introduces `testing/manual.csv` and https://github.com/w3c/wot-profile/pull/295 also.
Diffing them shows many differences...

danielpeintner:
@mlagally I think we need to have a _stable_ version of `manual.csv` otherwise it is not possible to provide any input...

benfrancis:
For https://github.com/w3c/wot-testing/pull/423 I've used the manual.csv from https://github.com/w3c/wot-testing/blob/main/events/2022.09.Online/Profile/manual.csv, which was updated around the same time as this PR so hopefully they're the same.

mlagally:
@danielpeintner, @mmccool, @k-toumura
Please use:
https://github.com/w3c/wot-testing/blob/main/events/2022.09.Online/Profile/manual.csv
and provide a PR with your response to https://github.com/w3c/wot-testing/events/2022.09.Online/Profile/Results/

PR #307, opened 2022-10-26
WIP: Initial draft of the APA checklist

mlagally:
Initial draft. 
Needs more consideration of accessibility requirements in the Profile.


mlagally:
APA review of TD spec was already done. The profile does not contain new a11y features.

Issue #313, opened 2022-11-03
Include the "auto" security scheme to the list of supported schemes


mlagally:

benfrancis:
@mmccool Do you remember what the rationale for this was?

Note that if the auto security scheme is added to Profiles, that would mean it has to be supported by _all_ compliant Consumers.

benfrancis:
See also the discussion in the related PR https://github.com/w3c/wot-profile/pull/314

benfrancis:
As of today PR #314 is still in draft and we've not heard any further justification for requiring that all Consumers support the `AutoSecurityScheme` in order to be conformant with any Profile.

Given our time constraints I suggest we don't add this constraint, and instead continue to require that conformant Things explicitly state security schemes in their Thing Description rather than rely on automatic negotiation. The `AutoSecurityScheme` feels like a nice to have, and something we could consider adding in 2.0.

@mmccool Please let us know within the next week if you still feel strongly about adding this, otherwise I suggest we should close PR #314 and remove the Profiles-1.0 label from this issue so it is no longer blocking the 2.0 release.

PR #314, opened 2022-11-03
Allow auto security scheme for other permitted security schemes

mmccool:
Resolves https://github.com/w3c/wot-profile/issues/313

1. Rather than simply adding "auto" to the list of supported schemes, which would have opened up negotiation of schemes NOT on the permitted list, I added an additional assertion that allows "auto" (and therefore security negotiation) but says only configurations consistent with the other permitted schemes is allowed
2. I'm not sure what the intent of assertion common-constraints-security-6 is.  
      - If it is for Discovery, then it overlaps with common-constraints-security-7 and is redundant.
      - If is not, and the intent is to allow security bootstrapping for other affordances, then "auto" should be used in the TD to indicate this.
      - Either way, I think it should be removed...

I will hold this PR in "draft" state until we resolve point 2.

### DIFF
--- a/index.html
+++ b/index.html
@@ -746,6 +746,13 @@
         <p><span class="rfc2119-assertion" id="common-constraints-security-3">
             A Thing MAY implement multiple security schemes.</span>
         </p>
+
+        <p><span class="rfc2119-assertion" id="common-constraints-security-auto">
+           The 
+           <a href="https://w3c.github.io/wot-thing-description/#autosecurityscheme"><code>AutoSecurityScheme</code></a>
+           MAY also be used, but only if negotation results in a security configuration
+           consistent with one of the <a href="#common-constraints-security-1">other permitted schemes</a>.</span>
+        </p>
         
         <p><span class="rfc2119-assertion" id="common-constraints-security-6">
           Conformant Consumers MUST support security bootstrapping for all 


benfrancis:
> I'm not sure what the intent of assertion common-constraints-security-6 is.
> 
> * If it is for Discovery, then it overlaps with common-constraints-security-7 and is redundant.

- common-constraints-security-6 is a requirement for Consumers
- common-constraints-security-7 is a requirement for Things

> * Either way, I think it should be removed, or perhaps changed to the following:
> 
> * ```
>     Conformant Consumers MUST support the <a href="https://w3c.github.io/wot-thing-description/#autosecurityscheme"><code>AutoSecurityScheme</code></a>  
>     and security negotiation for all <a href="#common-constraints-security-1">other permitted security schemes</a>.
>   </span></p>```
>   ```

I don't think so, "security negotiation" for interaction affordances (as per the `auto` security scheme used by Forms) is a different mechanism to "security bootstrapping" to gain access to a Thing Description.

> 3\. Should we restrict security bootstrapping for Discovery to the same set of permitted security schemes, i.e. restrict common-security-constraints-7 similarly?

That is the intent of the current assertions:
- "Conformant Consumers MUST support all of these security schemes."
- "Conformant Consumers MUST support security bootstrapping for all implemented security schemes"

I do acknowledge this is confusing though because it conflates "security schemes" in the general sense (applied to the authentication of a Thing Description) with "security schemes" in the specific sense of `SecurityScheme` metadata inside a Thing Description.

In general I think we need to more clearly distinguish between the security requirements for accessing a Thing Description vs. the security requirements of Forms. Perhaps separate sub-sections would help:

1. Thing Description Security
2. Form Security

mmccool:
Going to change this to non-draft.  Propose we merge this and discuss and resolve points 2 and 3 in new issues/PRs.

mlagally:
Arch call on Dec 7th:
Members on the call support to merge this unless we get an objection within a week, i.e. until Dec 14th.
McCool will create new issues for points 2 and 3 to continue the discussions separately.

benfrancis:
What actually needs to be implemented to support "automatic negotiation" for the BasicSecurityScheme and OAuth2SecurityScheme?

[Section 5.3.3.3](https://w3c.github.io/wot-thing-description/#autosecurityscheme) of the TD spec only provides a reference for Basic Authentication, and I'm not even sure it's linking to the right RFC.

mmccool:
Regarding Ben's question about implementation, negotiation is how HTTP *normally* works, so implementation is trivial (in fact required if you are implementing HTTP according to its spec).  What "auto" really means is that no information is provided in advance in the TD, it has to be discovered dynamically using normal HTTP mechanisms.  This is also to address a security consideration - there are some cases when you don't want to advertise the security mechanism used in advance (e.g. to discourage people scanning for vulnerabilities).

benfrancis:
As decided in the Profile call today and noted in #313, we will close this PR and re-visit this topic in 2.0.

PR #315, opened 2022-11-04
Improve link assertions

mlagally:


index.html
```diff
@@ -787,8 +787,16 @@ <h2>Links</h2>
 
       <p>
         <span class="rfc2119-assertion" id="common-constraints-links-1">The following keywords are defined for the
-          HTTP profiles and MUST be supported by all consumers.</span>
+          HTTP profiles and MAY be present in compliant TDs.</span>
```
sebastiankb, line 790: I'm not sure what "compliant" means here. Does also "non-compliant" TD exist?  Or do you mean "profile-enabled TDs"?

index.html
```diff
@@ -787,8 +787,16 @@ <h2>Links</h2>
...
+	<span class="rfc2119-assertion" id="common-constraints-links-1a">Compliant consumers that have the capabilities
```
sebastiankb, line 791: Again, I'm not sure what a "complaint consumer" means. Maybe you mean. "Consumers that only follows the HTTP Profile, have the capabilities...."

index.html
```diff
@@ -787,8 +787,16 @@ <h2>Links</h2>
 
       <p>
         <span class="rfc2119-assertion" id="common-constraints-links-1">The following keywords are defined for the
-          HTTP profiles and MUST be supported by all consumers.</span>
+          HTTP profiles and MAY be present in compliant TDs.</span>
+	<span class="rfc2119-assertion" id="common-constraints-links-1a">Compliant consumers that have the capabilities 
+	  to handle these relation types SHOULD treat the link targets appropriately.</span>
```
sebastiankb, line 792: I'm not sure why we force the consumer to do something with the relation type. E.g., if a user interface is implemented in such a way that it can also display an icon, it would be great to use this possibility provided in the links. However, if the UI is not supporting displaying an icon that should be also fine.

benfrancis:
This is a slight improvement in that it no longer says that it says all the keywords MUST be supported by all Consumers. However, the PR does not address all of the many issues I identified in #255.

I maintain that the only way to make this section implementable by all Consumers is to make all of the link relation types and media types optional, but they are already optional/recommended in the Thing Description specification which therefore makes this section redundant.

The Editor's Note should be a Note and there are grammatical errors introduced in this PR, but that's a bit of a moot point since I think the whole section needs to be removed.

_Maybe_ this section could _recommend_ a smaller set of link relation types than the TD specification (though what does that mean if a superset is already recommended in the TD spec?), and _recommend_ a set of media types (which the TD specification doesn't currently do), but that doesn't really guarantee out-of-the-box interoperability and any such list should be based on implementation experience, not just an arbitrary list.

benfrancis:
Also see https://github.com/w3c/wot-profile/issues/255#issuecomment-1305510762 for a potential alternative approach.

index.html
egekorkan: Agreeing with the comment above. Also, "handling appropriately" has no real meaning to be useful as an assertion.

index.html
```diff
@@ -787,8 +787,16 @@ <h2>Links</h2>
...
       </p>
+      <section class="ednote">
+      <p>For consumers that can render images or documents this implies to display appropriate documentation
```
egekorkan, line 795: I strongly suggest referring to https://github.com/w3c/wot-profile/issues/255#issuecomment-1305510762 for a real restriction on the rel values together with the media types.

egekorkan (changes requested):

mlagally:
Arch call on Nov 23rd:
Consider not to mandate but recommend "best effort"

index.html
mlagally: Should be "profile compliant" - changed.

index.html
mlagally: I have improved the language, please review the new text.

mlagally:
Arch call on Dec 7th:
Agree to merge.

Issue #138, opened 2021-11-18
generic UI support


mlagally:
The profile should contain enough information to build a generic UI.
Typical UIs for IoT applications contain text labels, gauges, timeline graphs.

benfrancis:
Similar to #136, we agreed [months ago](https://github.com/w3c/wot-profile/issues/73) that human readability is out of scope for the Core Profile, which is focused only on interoperability.

Are you proposing that we re-visit the scope of the specification again just over a month from our feature complete deadline?

This requirement and #136 are certainly not compatible.

We could re-visit whether human readability should be a goal of the Core Profile since it is at least not incompatible with an HTTP protocol binding, but I suggest we need group consensus in order to change direction again so late in the publication schedule.

benfrancis:
Following up on comments from other issues...

What exactly would it mean to make members like `title` and `description` mandatory? Would it mean that a WoT Consumer must stop parsing a Thing Description if these members are missing and not allow a user to interact with the device? That seems like a barrier to interoperability rather than an aid.

I'm not arguing that these strings are not important, of course they are important for usability and developers should absolutely be encouraged to include them in _all_ Thing Descriptions (not just those following the Core Profile). But imagine if a web browser refused to render a web page just because it had a missing `<title>` tag or `<meta name="description">` tag, or because it had too many levels of nested `<div>`s in the page*. Large parts of the web would just stop working.

The general approach on the World Wide Web is to make a best effort based on the content made available to the client and apply a series of fall-backs, not to just stop processing altogether if something that would be desirable to have is missing. There are other systems of incentives like search engine rankings and richer UI representations which encourage best practices.

For a web page, if no `<title>` element is provided for a page then browsers tend to fall back to displaying the URL.

For a Thing Description, if no `title` member is provided for an interaction affordance, then it can fall back to the interaction name provided as the key in the map.

I agree with the principle that Thing Descriptions should provide human-readable strings to help with the automatic construction of a UI, but refusing the interact with a device because it has a missing `description` member in one of its interaction affordances doesn't seem like the best way to incentivise that.

Instead, my suggestion is to propose a set of recommendations in the Thing Description specification which define best practices for _all_ Thing Descriptions, not make these constraints mandatory parts of an interoperability profile, since that could have the opposite than intended effect.

---

*Actually that last one doesn't sound like such a bad idea, maybe we should suggest it for the HTML specification 😂

benfrancis:
@mlagally wrote:
[...]

PR #316, opened 2022-11-04
Adding Common constraints for accessibility

mlagally:
Introducing a new section with two requirements to satisfy the needs of people with disabilities.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/316.html" title="Last updated on Nov 9, 2022, 1:08 PM UTC (95b435b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/316/f041378...95b435b.html" title="Last updated on Nov 9, 2022, 1:08 PM UTC (95b435b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -682,13 +682,13 @@
  	users with disabilities.
       </p>
       <p><span class="rfc2119-assertion" id="common-constraints-a11y-1">
-	  It is MANDATORY to provide a <code>title</code> that can be automatically rendered in a non-visual way 
-	  (e.g. using a screen reader),
+	  It is REQUIRED to provide a <code>title</code> that can be automatically rendered in a non-visual way 
+	  (e.g. using a screen reader)
 	  for things that may be used in deployments with users with disabilities.</span>
       </p>
       <p><span class="rfc2119-assertion" id="common-constraints-a11y-2">
 	  It is highly RECOMMENDED to provide a <code>description</code> that can be automatically rendered in a non-visual way 
-	  (e.g. using a screen reader),
+	  (e.g. using a screen reader)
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -674,6 +674,24 @@ <h1>Common Constraints</h1>
...
 
+    <!-- Accessibility -->
+
+    <section id="common-constraints-accessibilityy">
+      <h2 id="accessibility">Accessibility</h2>
+      <p>Authors of <em>Thing Descriptions</em> must ensure that the things described by them are accessible to
+ 	users with disabilities.
+      </p>
+      <p><span class="rfc2119-assertion" id="common-constraints-a11y-1">
+	  It is MANDATORY to provide a <code>title</code> that can be automatically rendered in a non-visual way
```
sebastiankb, line 685: I think, we should use here the rfc2119 terminology "SHOULD".

sebastiankb: does this apply to "all" title terms in the TD model or only the top level one?

sebastiankb (changes requested):

index.html
sebastiankb: fyi we have already this kind of discussion here: https://github.com/w3c/wot-profile/issues/138

egekorkan:
I want to record my opinion from the call:
It is not clear where these terms are required. If it is for Thing level, we already need title as required by the TD spec. If it is for all levels, then it would mean having these terms literally everywhere in the TD, including data schemas, which would make TDs explode. The assertions MUST be testable and these are currently not specific enough.

PR #318, opened 2022-11-09
Converting from ednote to note

mlagally:
Editor's notes are typically only in drafts.

The slightly modified text can go into the final spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/318.html" title="Last updated on Nov 9, 2022, 11:15 AM UTC (25ebb55)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/318/f041378...25ebb55.html" title="Last updated on Nov 9, 2022, 11:15 AM UTC (25ebb55)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1407,13 +1407,13 @@
             HTTP/1.1 204 No Content
           </pre>
         </section>
-        <section class="ednote">
+        <section class="note">
           <p>
-            The <code>readmultipleproperties</code> operation is currently excluded due to
+            The <code>readmultipleproperties</code> operation is  excluded due to
             the complexities of the request payload format and because it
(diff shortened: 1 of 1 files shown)


benfrancis:
This was actually intended as an editor's note which would be removed, but I don't object to keeping it in the specification if you want to.

mlagally:
> This was actually intended as an editor's note which would be removed, but I don't object to keeping it in the specification if you want to.

I would also agree to remove the entire note. Usually shorter text is better text ;-)

mlagally:
Arch call on Dec 7th:
Agree to merge

PR #325, opened 2022-11-15
Webhook subscription mechanism

mlagally:


mlagally:
Arch call on Nov 23rd:
- Align URL with URI wrt. TD
- DELETE does not take a payload
- Introduce a subscription id resource on which the unsubscribe operation is executed
- revert span->div

mlagally:
Created issue #329 to address the issue wrt. normative text for the entire spec.

mlagally:
Arch call on Dec 7th:
Text is an improvement, agree to merge.

index.html
```diff
@@ -3046,26 +3054,22 @@ <h4><code>subscribeevent</code></h4>
...
             <li>
-              A body with a subscription payload, serialized in JSON.
+              A body with a subscription request payload, serialized in JSON.
             </li>
-
-            <p class="ednote">
-              <!-- TODO --> TODO: Define subscription payload.
+            <p>
+              The subscription payload contains the URI for the event message listener
+              in the field with the <code>callbackURI</code> key.
```
benfrancis, line 3061: [Example 44](https://w3c.github.io/wot-profile/#example-44) uses `callbackURL` but this says `callbackURI`.

[Example 70](https://w3c.github.io/wot-thing-description/#example-70) in the TD spec also uses `callbackURL`.

index.html
```diff
@@ -3117,25 +3122,22 @@ <h4><code>unsubscribeevent</code></h4>
             the Web Thing with:
           </p>
           <ul>
-            <li>Method set to <code>POST</code></li>
-            <li>URL set to the URL of the <code>Event</code> resource</li>
+            <li>Method set to <code>DELETE</code></li>
+            <li>URL set to the URL of the <code>Event</code> resource followed by the <code>subscriptionId</code> of a valid subscription</li>
```
benfrancis, line 3126: [Example 44](https://w3c.github.io/wot-profile/#example-44) uses `subscriptionID`, but this says `subscriptionId`.

index.html
```diff
@@ -3085,12 +3089,13 @@ <h4><code>subscribeevent</code></h4>
               <code>Content-Type</code> header set to
               <code>application/json</code>
             </li>
-            <li></li>
+            <li>
+              <code>Accept</code> header set to <code>application/json</code>
+            </li>
```
benfrancis, line 3094: I don't think the `Accept` header is necessary here is it?

index.html
```diff
@@ -3117,25 +3122,22 @@ <h4><code>unsubscribeevent</code></h4>
...
-              object with a valid <code>subscriptionId</code>.
-            </li>
-            <p class="ednote">
-              <!-- TODO: define subscriptionId format - urn? -->
-              TODO: define subscriptionId format - urn?
-            </p>
           </ul>
+
+          <pre class="example" title="Unsubscribe using subscriptionId">
+            DELETE /things/lamp/events/overheated/1234-4544-1211 HTTP/1.1
```
benfrancis, line 3133: Assertion `http-webhook-profile-protocol-binding-subscribeevent-4b` says that the subscription ID should be sent in the payload, but here it is included in the URL. I agree it should be in the URL.

index.html
benfrancis: What should the response be? I would suggest `204 No Content`, as with the cancelaction protocol binding in the HTTP Basic Profile. This would also mean no `Accept` header is needed on this request.

index.html
```diff
@@ -3212,18 +3210,19 @@ <h4><code>subscribeallevents</code></h4>
               <code>Content-Type</code> header set to
               <code>application/json</code>
             </li>
+            <li>
+              <code>Accept</code> header set to <code>application/json</code>
+            </li>
```
benfrancis, line 3215: Responses don't need `Accept` headers.

index.html
```diff
@@ -3254,20 +3253,17 @@ <h4><code>unsubscribeallevents</code></h4>
...
-              object with a valid <code>subscriptionId</code>.
-            </li>
           </ul>
-          <p class="ednote">
-            <!-- TODO -->
-            TODO: define subscription id format - urn?
-          </p>
+
+            <pre class="example" title="Unsubscribe all using callbackURI">
+              DELETE /things/lamp/events HTTP/1.1
```
benfrancis, line 3264: The request URL should include the subscription ID.

PR #327, opened 2022-11-23
Implementation Report Update (indices, assertion text in template.csv)

mmccool:
- Updates report.html and associated csv files atrisk.csv and template.csv in testing
- Using new tooling as described in [TD PR #1752](https://github.com/w3c/wot-thing-description/pull/1752) to assign indices, order by appearance in the text, and include the assertion text in the autogenerated template.csv
- The recently added "other" template.csv that was in testing/inputs/template.csv (and before that in testing/inputs/results/template.csv) has been deleted to avoid confusion.  Please use the version in testing/template.csv as the basis of testing input files.

Note that the python script that previously generated template.csv did not deal with certain cases like table assertions, child assertions, and default assertions all of which are supported by the new tooling (even though these are not all used in the current version of the profile spec, it would be nice to be consistent).


mlagally (commented):
The implementation report has a section "automated validation results".
Are these results obtained automatically?

egekorkan:
They are in TD, not here. The same template results in that section appearing here.

mlagally:
> They are in TD, not here. The same template results in that section appearing here.

Why does this section contain assertions? Was there some automatic testing?

mlagally:
Approved in arch call on Nov 30

Issue #258, opened 2022-08-04
New section on Webhook message format


mlagally:
Create a new section that defines the message format for Webhook events as a replacement for the current CloudEvents.

egekorkan:
Before I make a PR about it, I would like to understand the requirements on the payload content for Webhook. I think that @benfrancis has more experience based on the comment at https://github.com/w3c/wot-profile/issues/126#issuecomment-1190420851 . I am copying his proposal here:

-     source (could be the URL of the TD, base URL, the URL of the event affordance, or the URL of the events meta interaction affordance): This is needed since the href is not necessarily unique for an event affordance in webhooks. Can be avoided if make it mandatory to have a separate href per event affordance
-     subject (event name, if not provided as part of source)
-     time
-     id (if an identifier separate from the timestamp is desired)
-     data

This is already less than what is proposed, which has fields like `specversion`, `contenttype` etc.

benfrancis:
Proposal:

| Member  | Type | Mandatory | Comment |
| ------------- | ------------- | ------------- | ------------- |
| `thingId`  | anyURI | Yes  | The ID of the Thing from which the event was emitted |  |
| `event` | string | Yes | The name of the event, taken from the key of its EventAffordance  |
| `id` | string | No | A unique identifier for the event, can be used when resuming a subscription. Can be any string including a UUID, URI or timestamp. |
| `timestamp` | dateTime | No | A timestamp for when the event took place |
| `data` | any | No | The event payload, matching the data schema of the EventAffordance |

E.g. 

```json
{
  "thingId": "https://mythingserver.com/things/mylamp1",
  "event": "overheated",
  "id": "223e3567-e89b-13d3-a456-426653",
  "timestamp": "2021-11-16T16:53:50.817Z"
  "data": 90
}
```

mlagally:
> This is already less than what is proposed, which has fields like `specversion`, `contenttype` etc. I think we need to better understand what is needed by a Webhook Consumer. I would also prefer to omit subject and source. The presence of `id` is needed to better argumented. What is the motivation behind having an identification mechanism for each event emission? I can invent some but a use case would be better.

id's are used to distinguish and order events when they have the same timestamp.

mlagally:
@benfrancis 
I like this proposal in principle, a few comments:

- [ ] The field name "event" is rather unspecific, suggest to call it "type" with content /events/<eventName>. In a similar way, property subscriptions should be /properties/<propertyName>
[...]

PR #330, opened 2022-11-30
WIP: cloud events payload

mlagally:
Repurposing the cloud events message format as an informative payload format for WebHooks.
This does not prevent other payload formats being defined and used.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3322,6 +3322,154 @@
     </section>
   </section>
 
+
+
+  <section id="sec-http-webhook-profile-message-format" class="informative">
+    <h3>Cloud Events Message Format</h3>
+    <p>
+      This section defines a JSON object format for events, which provides interoperability with cloud systems
+      that use the <a href="https://cloudevents.io/">cloud events specification</a> with the 
+      <a href="#sec-http-webhook-profile">HTTP Webhook profile</a>
+    <p>
+      This message format is one way of using the WebHook profile in a well defined way
+      that is supported by several cloud systems.
+    </p>
+    <p>
+      Using other message formats is not prohibited by this section - They should be used for deployments, 
+      where consumers use a different event payload.
+    </p>
+
+    <section id="http-webhook-profile-identifier">
+      <h2>Identifier</h2>
+      <p><span class="rfc2119-assertion" id="http-webhook-profile-1">
+          In order to denote that a given
+          <a href="https://www.w3.org/TR/wot-architecture/#dfn-thing">Web Thing</a>
+          conforms to the HTTP Webhook Profile and uses Cloud Events, 
+          its Thing Description MUST have a
+          <a href="https://w3c.github.io/wot-thing-description/#thing">
+            <code>profile</code></a> member  containing both profile identifiers:
+          <code>https://www.w3.org/2022/wot/profile/http-webhook/v1</code>
+          and <code>https://www.w3.org/2022/wot/profile/http-cloudevent/v1</code>
+          .</span>
+      </p>
+      <section class="note">
+        <p>Note that the <code>profile</code> member is an array that may contain multiple
+          profile entries, which indicates that a Web Thing conforms to all of the profiles in that array.</p>
+      </section>
+    </section>
+
+
+
+    <p class="rfc2119-assertion" id="http-webhook-profile-message-format-1">
+      If the consumer is providing an endpoint for cloud events, 
+      event notification messages SHOULD comply with the following data schema.
+    </p>
+
+    <p>
+      The format is aligned with the <a href="https://cloudevents.io/">cloud events specification</a>
+      to ensure interoperability with cloud systems that adopt that specification.
+    </p>
+    <p>
+      An event message contains a set of metadata properties and an optional data payload.
+      The metadata fields allow to determine the type and source of the event as well as
+      a timestamp indicating when the event occured and an optional data payload.
+    </p>
+    <p>
+      <span class="rfc2119-assertion" id="profile-json-event-payload">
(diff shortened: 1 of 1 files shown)


benfrancis:
In my opinion this should be a [payload binding template](https://w3c.github.io/wot-binding-templates/#payloads-intro) for the `application/cloudevents+json` media type, not a profile.

Profiles are meant to provide out-of-the-box interoperability for greenfield implementations, not compatibility with brownfield implementations.

Specifying multiple payload formats for the HTTP Webhook Profile will harm interoperability, not help it.

egekorkan:
I totally agree with @benfrancis . I have even proposed it at https://github.com/w3c/wot/blob/main/proposals/deliverable-proposals/binding-templates.md

It is a bad idea to propose a format for a greenfield application where the format is inherently incompatible with TDs (redundant fields, bad metadata/data separation). Plus, it is not a standard so we have to maintain it.

mlagally:
@egekorkan, @benfrancis  we have to be pragmatic on the fact that there is no binding template specification for cloud events. 
The proposed text does not preclude to define other payload formats, if needed, but if a deployment scenario already uses cloud events for notification messages, why not give guidance in this informative section ?

benfrancis:
I _strongly_ object to this being a profile for all the reasons already discussed.

I have proposed a prescriptive payload format for the HTTP Webhook Profile in https://github.com/w3c/wot-profile/issues/258#issuecomment-1216599450

egekorkan:
> if a deployment scenario already uses cloud events for notification messages, why not give guidance in this informative section ?

This is not a motivation for a profile. Also, if other payloads can be defined, what does a Profile consumer do? Try to implement parsers for all of them? How would the consumer distinguish the two payloads?

mlagally:
> I _strongly_ object to this being a profile for all the reasons already discussed.
> 
> I have proposed a prescriptive payload format for the HTTP Webhook Profile in [#258 (comment)](https://github.com/w3c/wot-profile/issues/258#issuecomment-1216599450)

@benfrancis 
We discussed the topic extensively in today's Profile call.

There were several supporters for defining two Webhook payload formats:
1. To drive WoT adoption in deployments to real world cloud systems that are already in the market and use CloudEvents 

2. A format for green field devices that can be implementing your proposal from [#258 (comment)](https://github.com/w3c/wot-profile/issues/258#issuecomment-1216599450)

The IoT market has evolved since multiple years and real world IoT cloud systems are deployed in the market. To keep the interoperability promise of WoT, we should not neglect but rather support these systems.

benfrancis:
We already have two separate HTTP-based profiles for delivering events and now you want to add a third?

I don't think we are on the same page about what profiles are for.

Profiles can only provide interoperability _within_ a profile, therefore the number of profiles needs to be as small as possible because the more profiles there are the more fragmented the Web of Things will be.

Ideally there would only be a single HTTP-based profile, but we couldn't agree on a single eventing mechanism so we agreed to have two separate profiles for events on the basis that SSE and Webhooks fit two different deployment models.

Now we can't agree on a single payload format for Webhooks so you're suggesting there should be two separate Webhook profiles. This will further fragment the Web of Things.

We need to choose a single payload format for Webhooks. @egekorkan and I have explained in detail why CloudEvents are not a good fit for the Web of Things data model, and that direct compatibility with brownfield systems is a non-goal for profiles.

We need to define a simple payload format for Webhooks, and it doesn't need to imitate CloudEvents, which is not even a standard.

If we keep splitting profiles every time we have a disagreement on the details then I'm afraid the whole concept of profiles is going to fall apart.

benfrancis:
@mlagally I've just read the [draft minutes](https://www.w3.org/2023/02/22-wot-profile-minutes.html) from the meeting, including @egekorkan's objections to your proposed direction which I note you didn't mention here.

Having read over the notes I'd like to again reiterate that there are currently two separate and complimentary approaches to interoperability on the Web of Things:
- Descriptive protocol binding templates for one-to-one integrations with brownfield systems
- Prescriptive profiles for plug-and-play interoperability between greenfield systems

Existing systems which use CloudEvents fall into the former category and are not a good fit for the latter category for all of the reasons already discussed (which Ege also reiterated in today's meeting). If you have an application which needs to integrate with existing systems using CloudEvents but can not be described using an existing binding template, then I suggest you contribute a binding template specifically for describing CloudEvents and don't try to shoehorn all of Oracle's requirements into the Profile specification.

mlagally:
@benfrancis 
> We need to choose a single payload format for Webhooks. @egekorkan and I have explained in detail why CloudEvents are not a good fit for the Web of Things data model, and that direct compatibility with brownfield systems is a non-goal for profiles.

If the Web of Things data model is not a good fit for CloudEvents, the TD specification has a bigger problem. 

I think we should not ignore the fact, that there are existing standards and de-facto standards (e.g. WebHooks), which came to the world to solve the eventing problem for large scale systems.

If we go down that route to require a specification of a dedicated (sub-) protocol each time we need a well-defined payload format (e.g. for action status responses, error responses etc.) we are fragmenting the problem space significantly.

@egekorkan The profile specification should be pragmatic, and I do not see a real problem with  parsing different event payload formats. A JSON parser can parse the object independently from the actual keys the object contains.

egekorkan:
> If the Web of Things data model is not a good fit for CloudEvents, the TD specification has a bigger problem.

We did not say that. TD can perfectly describe CloudEvents format. The problem is recommending it for a greenfield WoT application.

> Cloud Events are common ground in today's Cloud systems. Let's make sure we address these market needs with the WoT profiles specification.

That might be true (please provide references) but is this goal of the profile spec from my point of view.

Issue #41, opened 2020-10-05
Specific contraints on OAuth need to be defined


mlagally:

relu91:
Sadly right now, we only implemented client flow in node-wot (not sure if @farshidtz implemented other flows in its TDD). Therefore, I suggest to keep it simple by now and constraint possible OAuth flows to just the client.

farshidtz:
In our implementation, the TDD expects only the Access Token which can be obtained from any OAuth flow. The diagrams in [this comment](https://github.com/w3c/wot-discovery/pull/50#issuecomment-675614073) may be useful. Note that currently, we expect OpenID Connect's ID Token to also get the user claims (for user-oriented flows such as the code flow) and perform fine-grained access control. But we plan to drop OpenID Connect and use OAuth scopes for this matter.

egekorkan:
Just transferring my comments from #37 here:
- OAuth2 -> Should all scopes be supported?
- OAuth2 having all scopes/styles?

lu-zero:
Probably #331 addresses the issue already, shall we close it?

lu-zero:
Which ones? I'd keep what is proved to work and interoperate if possible.

egekorkan:
webthings implement code flow whereas node-wot implement client. So if there was decision like "let's pick what is already implemented", it was not correct. In any case, the decision should not be about what is already implemented but what is easier to implement and fulfils our requirements

PR #331, opened 2022-11-30
narrowing down oauth required flows

mlagally:


mmccool:
Actually, probably the client flow makes more sense.  We want one that can be used by automated systems and does not require a human in the loop.  Most of the flows do.

benfrancis:
For the record, WebThings Gateway currently implements the `code` flow, which does require user intervention. It enables third party web services to request access to one or more Things exposed by the gateway, with authorisation provided by a user who authenticates the request using their username and password.

I'm not very familiar with all the different OAuth2 flows because I haven't yet had to implement them myself in the context of WoT. With the `client` flow, how would a Consumer get authorisation to access a Thing it had never encountered before in an out-of-the-box interoperability type scenario? i.e. if a Consumer is given the URL of a new Thing Description (from some introduction mechanism), how does the Consumer then gain access to the Thing without user intervention? Does it require the owner of the Thing to be aware of the Consumer in advance?

mmccool:
See also [Issue #41](https://github.com/w3c/wot-profile/issues/41), which notes that node-wot only implements client.  It seems web-things only implements code.  We could allow both, but just say that a consumer needs to understand either but a producer/Thing is free to only use one (or none, if they use a different scheme altogether).

benfrancis:
@mmccool That sounds reasonable.

Putting aside what is already implemented, the reason for my question about the client flow is that I'd like to understand whether a Consumer needs advance knowledge about a Thing in order to authenticate with it, because if it does then that makes it less useful for the out-of-the-box interoperability use cases the profiles are intended to fulfil.

mlagally:
Arch call on Dec 7th:
Consensus to constrain Oauth for the next Working Draft to code and client flows.
Updated and merged the MR to reduce consumer implementation burden.

PR #332, opened 2022-11-30
Testing CSV cleanup

mmccool:
- get rid of extra whitespace in assertion text in template.csv
- copy template.csv to manual.csv (everything is manually tested at present)
- note that manual.csv is needed for Dec 2022 testfest
- regen implementation report: now all assertions are in one table


mlagally:
Arch call on Dec 7th:
Important for the testfest. Agree to merge.

PR #340, opened 2022-12-09
Prepare 2-WD Release (phase 3)

mmccool:
Fix additional errors
- remove reference to wot-security-best-practices
- fix typos in urls and fragment identifiers (missing TR in one URL, sometime sec-security-considerations is plural, sometimes not...)

Still some errors but we can ignore.  For example, the event-source reference has a "permanent" redirect to whatwg but this is being migrated back to W3C, and the link does work, so it's not really an error.  Likewise there is a self-link to the new URL of this document which is broken since it's not yet published.

Merging immediately so I can run link checker again...

### DIFF
--- a/publication/2-wd/Overview.html
+++ b/publication/2-wd/Overview.html
@@ -2428,7 +2428,7 @@ is set to <code>true</code> then the Web Thing <em class=
 <p><span class="rfc2119-assertion" id=
 "http-basic-profile-protocol-binding-invokeaction-8">If the
 <code>synchronous</code> member of the <a href=
-"https://www.w3.org/wot-thing-description11/#actionaffordance"><code>
+"https://www.w3.org/TR/wot-thing-description11/#actionaffordance"><code>
 ActionAffordance</code></a> [<cite><a class="bibref"
 data-link-type="biblio" href="#bib-wot-thing-description11" title=
 "Web of Things (WoT) Thing Description 1.1">wot-thing-description11</a></cite>]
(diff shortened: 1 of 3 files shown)


mmccool:
Reran, no new errors found.   Going to consider link checking complete.  I AM getting a "forbidden" error for ev.buaa.edu.cn which is Beihang, but this was added by the ReSpec exporter.   Here is the complete list of remaining errors:

```
error Line: 282 https://www.w3.org/TR/2022/WD-wot-profile-20221209/
Status: 404 Not Found
The link is broken. Double-check that you have not made any typo, or mistake in copy-pasting. If the link points to a resource that no longer exists, you may want to remove or fix the link.

error Line: 344 https://ev.buaa.edu.cn/
Status: 403 Forbidden
The link is forbidden! This needs fixing. Usual suspects: a missing index.html or Overview.html, or a missing ACL.

info Line: 318 mailto:public-wot-wg@w3.org?subject=%5Bwot-profile%5D%20YOUR%20TOPIC%20HERE
Status: (N/A) Access to 'mailto' URIs has been disabled
Accessing links with this URI scheme has been disabled in link checker.

warning Lines: 3635, 4825, 4827 https://www.w3.org/TR/eventsource/ redirected to https://html.spec.whatwg.org/multipage/
Status: 301 -> 200 OK
This is a permanent redirect. The link should be updated to point to the more recent URI.

Broken fragments:
https://www.w3.org/TR/eventsource/#the-eventsource-interface (line 3635)
```
For the last "Broken fragment", I have checked that https://www.w3.org/TR/eventsource/#the-eventsource-interface does in fact load the correct page.

PR #342, opened 2022-12-09
Prepare 2-WD Release (phase 5)

mmccool:
Fix pub rules warnings
- update in-text direct links to tools.ietf to use rfc-editor instead
- update link in references to use datatracker instead of tools.ietf

These also avoid some redirection warnings in the link checker.

Need to merge, then will re-run both pubrule and link checker.


mmccool:
Pub rules now does not flag tools.ietf as a problem, and only has two other ignorable warnings - "Use Link Checker", and "Validate CSS"

mmccool:
Link checker also no longer complains about a redirect from tools.ietf

PR #343, opened 2022-12-13
Implementation Report Update - Dec 2022

mmccool:
Include new results from Dec 2022 Testfest.  Includes Consumer/Thing/TD categories.


ashimura:
[rendered version by Statically](https://cdn.statically.io/gh/w3c/wot-profile/88f9f7385ae52aa4ec68505450cd21bb33e9e664/testing/report.html)

mmccool:
So it is odd that there are no Result files from Oracle, although there are TDs.  Suggest waiting until CSV Result files are submitted to wot-testing for the Oracle implmentations, and then I will update this report and merge this PR.

mlagally:
Arch call on Dec 21st.: Agree to merge

PR #346, opened 2022-12-21
Aadd clarification on server behavior of webhook consumer

### LINKED ISSUE #322: Brief section to describe a Consumer
We use the term "Consumer" in many areas of the spec without specifically describing it.
Recent conversations showed that we need additional clarifications.

We need to add a brief section that describes Consumers with WebHooks

### LINKED ISSUE #328: Improving webhook section to mention that Consumers need to have HTTP server function
#319 I have mentioned this but did not have time to add this in that PR. We should mention that the consumer of a Webhook Profile compliant need to have an HTTP server function

egekorkan:
fixes #322 
fixes #328


mlagally:
Profile call on Jan 11th: Agree to merge

PR #362, opened 2023-02-08
Finalize Explainer.md

mlagally:


mlagally:
Arch call on Feb. 8th:
Agree to merge, we can improve with dedicated issues and PRs.

PR #364, opened 2023-02-13
Security http sections

lu-zero:
Initial wording for #6.

***
<a href="https://pr-preview.s3.amazonaws.com/luminem/wot-profile/pull/364.html" title="Last updated on Feb 22, 2023, 12:28 PM UTC (5bbd014)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/364/b756329...luminem:5bbd014.html" title="Last updated on Feb 22, 2023, 12:28 PM UTC (5bbd014)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -643,11 +643,14 @@
           </ul>
         </div>
         <p><span class="rfc2119-assertion" id="common-constraints-security-2">
-            Conformant Consumers MUST support all of these security schemes.</span>
+            Conformant Consumers MUST support at least all of these security schemes.</span>
         </p>
 
         <p><span class="rfc2119-assertion" id="common-constraints-security-3">
             A Thing MAY implement multiple security schemes.</span>
+        </p>
+        <p><span class="rfc2119-assertion" id="common-constraints-security-4">
+            A Thing MUST support at least one of the above security schemes.</span>
         </p>
 
         <p><span class="rfc2119-assertion" id="common-constraints-security-6">
--- a/index.html
+++ b/index.html
@@ -901,7 +901,7 @@
 
     <!-- Errors -->
     <section id="common-constraints-errors">
-      <h2 id="error-responses">Errors</h2>
+      <h3 id="error-responses">Errors</h3>
       <p>
         <span class="rfc2119-assertion" id="common-constraints-errors-1">
           If any of the operations defined in the protocol bindings of HTTP


lu-zero:
@benfrancis I hope I addressed your feedback.

benfrancis:
I'm afraid I don't agree with this change. Please see https://github.com/w3c/wot-profile/issues/6#issuecomment-1428472260

I don't think any of the assertions in Common Constraints can be assumed to be applicable to all future profiles, only the current set of HTTP profiles. The current Common Constraints section _is_ an HTTP Common Constraints section.

I also don't see why some of the constraints you've moved to the separate HTTP Common Constraints section could not apply to other protocols in the future like CoAP.

I suggest leaving this as it is and re-assessing at such a point as we add non-HTTP Profiles. Only then will a separate section be necessary.

lu-zero:
60ef6cc7b4dbe4139cc90c827191694f95a5e2a7 can be omitted from the patchset without additional changes, is the rest fine for you?

index.html
```diff
@@ -643,24 +643,31 @@ <h2>Security</h2>
...
         </div>
         <p><span class="rfc2119-assertion" id="common-constraints-security-2">
-            Conformant Consumers MUST support all of these security schemes.</span>
+            Conformant Consumers MUST support at least all of these security schemes.</span>
         </p>
-	    
+
         <p><span class="rfc2119-assertion" id="common-constraints-security-3">
-            A Thing MAY implement multiple security schemes.</span>
+            A Thing MAY implement multiple security schemes, at least one of the above MUST be supported.</span>
```
benfrancis, line 650: As I understand it the tooling for extracting assertions doesn't support multiple RFC 2119 key words in a single assertion. You probably need to split this into two separate assertions with their own id.

index.html
```diff
@@ -643,24 +643,31 @@ <h2>Security</h2>
...
         </p>
-        
+
+        <p>
+            The <a href="https://w3c.github.io/wot-thing-description/#basicsecurityscheme">
+                <code>BasicSecurityScheme</code>
+            </a> SHOULD be avoided if the underlying transport is not
+            <a href="https://w3c.github.io/wot-architecture/#sec-security-consideration-secure-transport">secure</a>.
```
benfrancis, line 657: Whilst I don't disagree with this assertion (where it's possible to use TLS), I'm not sure this advice is unique to the Basic security scheme. A man-in-the-middle attack can be used to intercept tokens in the OAuth2 security scheme too.

Note that the [Security Considerations](https://w3c.github.io/wot-profile/#security-considerations) section already says:
> The security considerations of the [WoT Architecture](https://w3c.github.io/wot-architecture/#sec-security-consideration) and [WoT Thing Description](https://w3c.github.io/wot-thing-description/#sec-security-considerations) MUST be adopted by compliant implementations.

This includes WoT Architecture's [section](https://w3c.github.io/wot-architecture/#sec-security-consideration-secure-transport) on secure transport which you linked to, which goes into more detail and mentions that tokens are at risk as well as passwords.

Also note that assertions need to be wrapped in a `<div>` or a `<span>` with `class="rfc2119-assertion"` and a unique `id`.

mmccool:
Let's discuss this further in the Security call next week.  I would like to clean up this PR by next week's Profile call.  @benfrancis thanks for your review, please look for an update next week that I hope will address your concerns (may be a new PR, we'll see if this one can be repaired first).

index.html
egekorkan: Some comments:
- We can change to MUST in the profile ?
- TLS should be for all security mechanisms indeed (except nosec).

egekorkan:
It is not here anymore but I think that each profile should explain the security scheme on its own. Is OAuth2 usable in Webhook?

lu-zero:
Hopefully now it is more acceptable.

index.html
benfrancis: > We can change to MUST in the profile ?

Regular reminder: TLS doesn't work on local networks. This would effectively forbid the use of profiles on local networks.

Use case: WebThings Gateway currently exposes its API via http://gateway.local, and _optionally_ https://{subdomain}.webthings.io, and uses profiles on both.

index.html
```diff
@@ -643,24 +643,25 @@ <h2>Security</h2>
...
         <p><span class="rfc2119-assertion" id="common-constraints-security-2">
-            Conformant Consumers MUST support all of these security schemes.</span>
+            Conformant Consumers MUST support at least all of these security schemes.</span>
         </p>
-	    
+
         <p><span class="rfc2119-assertion" id="common-constraints-security-3">
-            A Thing MAY implement multiple security schemes.</span>
+            A Thing MAY implement multiple security schemes</span>,
+            <span class="rfc2119-assertion" id="common-constraints-security-4"> at least one of the above MUST be supported.</span>
```
benfrancis, line 651: I suggest splitting this into two sentences. Once these assertions get extracted out into implementation reports the latter is not going to make any sense.

benfrancis:
@egekorkan wrote:
> Is OAuth2 usable in Webhook?

That's a good question, I don't know.

index.html
mlagally: I agree in principle to improve the language, however splitting into two sentences does not solve the problem. The 2nd sentence should contain the members of the list above.

index.html
```diff
@@ -643,12 +643,15 @@ <h2>Security</h2>
           </ul>
         </div>
         <p><span class="rfc2119-assertion" id="common-constraints-security-2">
-            Conformant Consumers MUST support all of these security schemes.</span>
+            Conformant Consumers MUST support at least all of these security schemes.</span>
```
danielpeintner, line 646: Do you mean...

Consumers MUST support at least **one** of these security schemes?

IF not.. "at least all" sounds strange, then I would rather stick to "all"

benfrancis: Consumers must support all of the security schemes in order to guarantee interoperability with all Things using the profile.

I agree the existing wording made more sense.

mlagally: @danielpeintner 
We discussed it in today's call and agreed to merge it. If there's a need to improve the text, please create a new issue.

index.html
```diff
@@ -643,12 +643,15 @@ <h2>Security</h2>
...
         </p>
 
         <p><span class="rfc2119-assertion" id="common-constraints-security-3">
             A Thing MAY implement multiple security schemes.</span>
         </p>
+        <p><span class="rfc2119-assertion" id="common-constraints-security-4">
+            A Thing MUST support at least one of the above security schemes.</span>
```
benfrancis, line 653: I'm sorry for asking to keep asking for further changes but there are now two similar sounding assertions which say different Things:

> Below is a list of security schemes which conformant Web Things MAY use

> A Thing MUST support at least one of the above security schemes.

The text originally said "Below is a list of security schemes which conformant Web Things MAY use and conformant Consumers MUST support" but I believe it was split into two assertions because sentences can't include two RFC 2119 keywords.

The former assertion also uses the term "MAY" incorrectly, which is my fault.

Let's just stop dancing around grammatical structures and duplicate the list 🙂...

Conformant Web Things MUST use at least one of the following security schemes [[wot-thing-description]]:
- NoSecurityScheme
- BasicSecurityScheme
- OAuth2SecurityScheme with the code or client flow.

Conformant Consumers MUST support all of the following security schemes [[wot-thing-description]]:
- NoSecurityScheme
- BasicSecurityScheme
- OAuth2SecurityScheme with the code or client flow.

benfrancis (changes requested):

Issue #299, opened 2022-10-04
Contradiction with the Arch spec on protocol bindings


egekorkan:
There is the following assertion in the arch spec at https://w3c.github.io/wot-architecture/#hypermedia-driven :

> [Protocol Bindings](https://w3c.github.io/wot-architecture/#dfn-wot-protocol-binding) MUST be serialized as [hypermedia controls](https://w3c.github.io/wot-architecture/#sec-hypermedia-controls) to be self-descriptive on how to activate the [Interaction Affordance](https://w3c.github.io/wot-architecture/#dfn-interaction-affordance).

However, the profiles contradict this assertion since they define a protocol binding inside the specification and this is referred to only by the `profile` term in the top level, and thus out of the hypermedia controls.

benfrancis:
See also: https://github.com/w3c/wot-thing-description/issues/1674

I have filed https://github.com/w3c/wot-architecture/issues/850

egekorkan:
If all the TDs look like the ones submitted by @mlagally to testfest, we should be ok: https://github.com/w3c/wot-testing/blob/main/data/input_2022/Profile/TD/Oracle/TDs/BluePump%20WebThing.td.jsonld

mlagally:
Profile call on March 3rd: 
Needs more discussion and a concrete proposal.

Alternatives:
1. remove normative language from Architecture before REC
2. Add language to require "complete forms"

Revisit in Profile call on March 29th

benfrancis:
Defining the protocol binding declaratively in every Form in addition to specifying it in the profile would negate most of the benefits of using a profile, which allow for much simpler Thing Descriptions by having Consumers assume a wider range of defaults. It's also not possible to describe many aspects of the profile protocol bindings declaratively in forms.

The TD and Architecture specifications should be updated to say that profiles can also define protocol bindings, as I suggested in https://github.com/w3c/wot-thing-description/issues/1674 and https://github.com/w3c/wot-architecture/issues/850, since this is what is currently done.

I hope that we can improve this situation in 2.0 by having the profiling mechanism and binding mechanism work together, see https://github.com/w3c/wot-profile/issues/285#issuecomment-1433087304

benfrancis:
[...]

PR #368, opened 2023-03-01
Add TD assertion conformance

### LINKED ISSUE #349: Are all TD 1.1 assertions valid in all profiles?
This is a question for the profiling mechanism in general. Since profile-compliant Things generate TDs, this should mean that they have to comply with all the assertions in the TD spec. For example, the following assertion is in the TD spec:

> When title and titles or description and descriptions are present in a TD document, each title and description member SHOULD have a corresponding titles and descriptions member, respectively.

My answer to my question is simply yes but I think that an assertion is needed in the profile spec. In the arch spec, the section https://w3c.github.io/wot-architecture/#profile-description-methodology says 

> These constraints define a subset of valid [WoT Thing Descriptions](https://w3c.github.io/wot-architecture/#dfn-wot-thing-description) by defining additional rules on various aspects of the [WoT Thing Description](https://w3c.github.io/wot-architecture/#dfn-wot-thing-description)

However, that is not an assertion neither and a valid TD is not a definition that is strictly defined, see:

> Formally, a valid TD satisfies all the assertions in this specification, but not all assertions can be validated given only the JSON serialization, for instance, the assertions listed under [8. Behavioral Assertions](https://w3c.github.io/wot-thing-description/#behavior) that relate a TD to the behavior of a Thing that it describes. Extensions are also problematic, in that even if formal metadata is given for validating an extension, dynamically fetching this metadata in a deployment might pose a privacy risk. In this section, therefore, we name and define various levels of validation appropriate for different contexts.

egekorkan:
fixes #349

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/368.html" title="Last updated on Mar 22, 2023, 12:05 PM UTC (5484a03)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/368/439d590...5484a03.html" title="Last updated on Mar 22, 2023, 12:05 PM UTC (5484a03)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -520,7 +520,7 @@
         specification. [[wot-thing-description11]].
       </span>
       <span class="rfc2119-assertion" id="profiling-mechanism-5">
-        All Things and Consumers conforming to a profile MUST first satisfy the assertions
+        All Things and Consumers conforming to a profile MUST satisfy the assertions
 	specified in the [[wot-thing-description11]].
       </span>
     </p>
--- a/index.html
+++ b/index.html
@@ -519,6 +519,10 @@
         the document is using version 1.1 of the Thing Description
         specification. [[wot-thing-description11]].
       </span>
+      <span class="rfc2119-assertion" id="profiling-mechanism-5">
+        All Things and Consumers conforming to a profile MUST first satisfy the assertions
+	specified in the [[wot-thing-description11]].
+      </span>
     </p>
     <pre class="example">
     {
--- a/index.html
+++ b/index.html
@@ -520,8 +520,9 @@
         specification. [[wot-thing-description11]].
       </span>
       <span class="rfc2119-assertion" id="profiling-mechanism-5">
-        All Things and Consumers conforming to a profile MUST satisfy the assertions
-	specified in the [[wot-thing-description11]].
+        All Things and Consumers conforming to a profile MUST satisfy the assertions specified in the [[wot-thing-description11]],
+        except for the assertion <code>td-context-ns-td10-namespacev10</code> with text <code>TD 1.1 consumers MUST accept 
+          TDs satisfying the W3C WoT Thing Description 1.0 [wot-thing-description] specification.</code>
       </span>
     </p>
     <pre class="example">


danielpeintner (approved):
LGTM.
What struggles me a bit is the use of "first". It reads to me as if there would be an "order" of assertions. Maybe skip it?
Anyhow, fine for me either way...

benfrancis (changes requested):
I did a very quick review of all the mandatory assertions in the TD specification to make sure this doesn't have any unintended side effects.

There are three assertions which raised questions for me. I think probably only the first one is a problem that needs addressing.

> TD 1.1 consumers MUST accept TDs satisfying the W3C WoT Thing Description 1.0 [[wot-thing-description](https://w3c.github.io/wot-thing-description/#bib-wot-thing-description)] specification.

This is problematic for Profile conformant Consumers because there was no `profile` member in Thing Description 1.0. If a Profile conformant Consumer also has to support Thing Description 1.0 then it also needs to support TDs where none of the assumptions implied by profiles apply. Implementation complexity therefore explodes and and interoperability goes out of the window. **I suggest we need to add an assertion to exempt Profile conformant Consumers from this TD assertion.**

> If no response name-value pair is provided, it MUST be assumed that the content type of the response is equal to the content type assigned to the Form instance

One slight subtlety here is that in the HTTP SSE Profile we assume the `text/event-stream` response type from the `subprotocol` member of the Form being set to `sse`. Given this is fixed in the [SSE specification](https://html.spec.whatwg.org/multipage/server-sent-events.html), I think that's a reasonable assumption.

> If the content type of an additional expected response differs from the content type of the form, the Form instance MUST include an entry in the array associated with the name additionalResponses that includes a value for the name contentType. If the data schema of an additional expected response differs from the output data schema of the interaction, the Form instance MUST include an entry in the array associated with the name additionalResponses that includes a value for the name schema.

WoT Profile has an assertion which says that if a body is included in an error response, it MUST conform with the Problem Details Format [RFC7807]. If applied strictly, the above assertions mean that every affordance which may provide an error response must provide `additionalResponses` objects for every possible error, and if that error response provides a body then it must also denote the `application/problem+json` content type. Hopefully the data schema would be implied by the content type otherwise every error response would also have to include a data schema for the problem details format. This all seems a bit excessive to me, but I guess it's up to implementations how many possible error states they want to describe.

egekorkan:
@benfrancis I think I agree with your points and that we can add the exception for the first one. The other two, I do not know yet if they need additional assertive text in the profile.

index.html
```diff
@@ -519,6 +519,10 @@ <h2>Profiling Mechanism</h2>
         the document is using version 1.1 of the Thing Description
         specification. [[wot-thing-description11]].
       </span>
+      <span class="rfc2119-assertion" id="profiling-mechanism-5">
+        All Things and Consumers conforming to a profile MUST first satisfy the assertions
```
mlagally, line 523:
```suggestion
        All Things and Consumers conforming to a profile MUST satisfy the assertions
```

mlagally: There's one requirement that needs more consideration:

TD Spec section 5.3.1.1:

> TD 1.1 consumers MUST accept TDs satisfying the W3C WoT Thing Description 1.0 [[wot-thing-description](https://www.w3.org/TR/wot-thing-description11/#bib-wot-thing-description)] specification.

Since the Profile contains additional constraints and targets to simplify the implementation, we should drop this requirement in the Profile, or change it to RECOMMENDED in the TD spec.

mlagally (changes requested):

egekorkan:
@mlagally I have added the exception.

@benfrancis , if we add text proposed at https://github.com/w3c/wot-profile/issues/299#issuecomment-1479406220 , we do not need any text about this in the profile.

mlagally:
Profile call on April 5th: Agree to merge - Please continue discussion on issue #299 separately.

Issue #369, opened 2023-03-01
Clarifications on async actions


mlagally:
Discussion on Profile call on March 1st.:
Support for asynchronous actions is necessary, but need further clarifications:

Actions do not have a way to subscribe, i.e. no eventing mechanism that tells you about a the action state. There could be an event mechanism.

Multiple invocations of actions need further clarification, does a new invocation cancel the previous one?

If invocations can queue up, this must be signaled (to the consumer)
Also when the queue is full, there must be a way to communicate that.

Current implementation where a new invocation cancels the previous invocation works, i.e. no queue, works.

A queueing mechanism can consume resources that go beyond the space of a constrained device.

benfrancis:
> Actions do not have a way to subscribe, i.e. no eventing mechanism that tells you about a the action state. There could be an event mechanism.

I agree this would be useful, but there is currently no operation type in the Thing Description specification that would cover this. I filed https://github.com/w3c/wot-thing-description/issues/1775 last week, but any new operations will not be added until Thing Description 2.0, which is after the target publication date for WoT Profile 1.0.

> Multiple invocations of actions need further clarification, does a new invocation cancel the previous one?

No. There is an explicit `cancelaction` operation, and nowhere does it say that invoking an action must cancel another one or that multiple action instances can not be in the `running` state at the same time.

> If invocations can queue up, this must be signaled (to the consumer)

The `pending` value of the [ActionStatus](https://w3c.github.io/wot-profile/#ActionStatus) object indicates that an action invocation request has been received but the action has not yet started.

Note that the current asynchronous actions protocol binding is not explicitly a queue, there's nothing to say that actions can not be performed in parallel. Whether more than one action can be performed at a time may depend on the individual affordance, e.g. it's not possible to move a robot arm in two directions at the same time. I don't think whether actions are queued or performed in parallel is something that can be fixed in the profile, it is implementation specific.

> Also when the queue is full, there must be a way to communicate that.

I would expect an error response to an invokeaction request (e.g. 503 Service Unavailable). This could be made explicit in an assertion.

> Current implementation where a new invocation cancels the previous invocation works, i.e. no queue, works.

I would consider this to be implementation specific (i.e. down to an individual Thing). That pattern might work well for a robotic arm, but not for a printer queue for example.

> A queueing mechanism can consume resources that go beyond the space of a constrained device.

This is true of may operations, not just asynchronous actions. Again, I would expect an error response in a situation like this.

lu-zero:
[...]

PR #371, opened 2023-03-02
Async action clarifications

lu-zero:
Addressing part of #369.

***
<a href="https://pr-preview.s3.amazonaws.com/luminem/wot-profile/pull/371.html" title="Last updated on Mar 2, 2023, 1:46 PM UTC (2b3f6eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/371/439d590...luminem:2b3f6eb.html" title="Last updated on Mar 2, 2023, 1:46 PM UTC (2b3f6eb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1671,7 +1671,19 @@
               "timeRequested": "2021-11-10T11:43:19.135Z"
             }
           </pre>
-        </section>
+          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-22">
+          <p>
+            In resource constrained environments, if the buffer becomes full then a Web Thing SHOULD
+            remove the oldest Action result ActionStatus from its buffer to make room for a newer one.
+          </p>
+          </div>
+          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-23">
+          <p>
+            A Web Thing MUST return a <code>503</code> error response if the invocation cannot be enqueued,
+            e.g. the internal pending queue is full.
+          </p>
+          </div>
+         </section>
 
         <section id="http-basic-profile-protocol-binding-queryaction">
           <h5><code>queryaction</code></h5>
--- a/index.html
+++ b/index.html
@@ -1673,14 +1673,13 @@
           </pre>
           <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-22">
           <p>
-            In resource constrained environments, if the buffer becomes full then a Web Thing SHOULD
-            remove the oldest Action result ActionStatus from its buffer to make room for a newer one.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1670,7 +1671,19 @@ <h6 id="async-action-response">Asynchronous Action Response</h6>
               "timeRequested": "2021-11-10T11:43:19.135Z"
             }
           </pre>
-        </section>
+          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-22">
+          <p>
+            In resource constrained environments, if the buffer becomes full then a Web Thing SHOULD
+            remove the oldest Action result ActionStatus from its buffer to make room for a newer one.
```
benfrancis, line 1677: On reflection the term "buffer" implies certain implementation details about the data structure which may not apply to all implementations. I would suggest a more generic term like "list", since in the API the collection of action requests is represented as a list/array of `ActionStatus` objects. There are many ways this could be modelled in different programming languages on the back end.

Also, it should be made clear that only completed/failed actions should be removed to make space. How to prioritise older vs. newer pending or running action requests needs to be application specific.

Also, not every `ActionStatus` object will be the same size, so it may be necessary to remove more than one old object to make room for one new object.

```suggestion
            In resource constrained environments, the ActionStatus objects of older completed/failed actions MAY be deleted to make room for newly invoked actions.
```

To be honest, I think what remains is already covered by the [Note](https://w3c.github.io/wot-profile/#h-note-3) at the end of the section.

index.html
```diff
@@ -1670,7 +1671,19 @@ <h6 id="async-action-response">Asynchronous Action Response</h6>
...
+          </p>
+          </div>
+          <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-23">
+          <p>
+            A Web Thing MUST return a <code>503</code> error response if the invocation cannot be enqueued,
+            e.g. the internal pending queue is full.
```
benfrancis, line 1683: This might need to be more specific, because there are other reasons an action invocation may not be accepted which might better fit a different response code like `400`, `401`, `403`, `404` or `500`. 

```suggestion
            A Web Thing SHOULD return a <code>503</code> error response if the invocation cannot be accepted because the action is unavailable,
            e.g. because the Thing is overloaded.
```

benfrancis (changes requested):
I think this is good, but some terms are a bit too implementation specific and the circumstances that would lead to a 503 error not specific enough.

index.html
lu-zero: Agreed, thank you for the review.

benfrancis (approved):

PR #374, opened 2023-03-06
Define values of in and name for basic security scheme

mmccool:
In Issue #6  we noted some missing details around the parameters for the "basic" security scheme that this PR tries to address:
- "in" should be "header".  As this is the default value, you can just leave it out.  This PR adds an assertion saying you can either leave it out or use the default value of "header".
- The value for "name" is trickier.  First, it does not have a default value, but is optional, which frankly is an oversight; it means it is technically undefined if not given.  However, in the HTTP specification it also has two possible values, whether or not a proxy is used.
- This PR adds two assertions for "name", one for when a proxy is used, one for when it is not.  For compatibility with the TD spec while avoiding undefined terms, this also means that "name" must ALWAYS be given.  An alternative would be to *define* a default value for it using these rules.

We also haven't said anything about proxies in the profiles spec, but the current text does not disallow them that I can see.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-profile/pull/374.html" title="Last updated on Mar 6, 2023, 2:49 PM UTC (196805f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/374/439d590...mmccool:196805f.html" title="Last updated on Mar 6, 2023, 2:49 PM UTC (196805f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -652,6 +652,16 @@
         <p><span class="rfc2119-assertion" id="common-constraints-security-4">
             A Thing MUST support at least one of the above security schemes.</span>
         </p>
+        <p><span class="rfc2119-assertion" id="common-constraints-security-basic-in">
+            For the <code>BasicSecurityScheme</code> the "in" field MUST be either omitted
+            or be given its default value of "header" as defined in [[wot-thing-description11].</span>
+           <span class="rfc2119-assertion" id="common-constraints-security-basic-name">
+            For the <code>BasicSecurityScheme</code> the "name" field MUST be provided using
+            the value "Authorization" if a "proxy" endpoint is not given.</span>
+           <span class="rfc2119-assertion" id="common-constraints-security-basic-proxy">
+            For the <code>BasicSecurityScheme</code> the "name" field MUST be provided using
+            the value "Proxy-Authorization" if a"proxy" endpoint is given.</span>
+        </p>
 
         <p><span class="rfc2119-assertion" id="common-constraints-security-6">
           Conformant Consumers MUST support security bootstrapping for all


index.html
```diff
@@ -652,6 +652,16 @@ <h2>Security</h2>
         <p><span class="rfc2119-assertion" id="common-constraints-security-4">
             A Thing MUST support at least one of the above security schemes.</span>
         </p>
+        <p><span class="rfc2119-assertion" id="common-constraints-security-basic-in">
+            For the <code>BasicSecurityScheme</code> the "in" field MUST be either omitted
+            or be given its default value of "header" as defined in [[wot-thing-description11].</span>
```
benfrancis, line 657: I appreciate that this simplifies implementation for Consumers. However, as I mentioned in #6, an unfortunate side effect of this is that it will make it impossible to use the `BasicSecurityScheme` with the Server-Sent Events and WebSocket APIs in web browsers, since they provide no mechanism to set an `Authorization` header via these APIs.

That also effectively excludes the `BasicSecurityScheme` from being used by the HTTP SSE Profile in web application implementations.

Is this an intentional constraint?

I don't personally have a use for `BasicSecurityScheme` in these circumstances because WebThings uses JWT Bearer tokens (which for WebSockets and SSE have to be provided in a query string by the web front end), but I can imagine that some Thing implementations might prefer the simpler `BasicSecurityScheme` if the only alternative is a full OAuth2 implementation.

mmccool: I'm not sure we WANT to use basic for SSE.  To be clear, these defaults correspond to what HTTP *normally* does for BasicAuth, so no changes to implementations should be necessary.  Changes to the TD are only needed to describe what the protocol is doing.  To be honest, I will see if we can fix the issue with "name" in the TD spec prior to PR, although the variant default values for proxy/non-proxy makes it annoying.

The *right* thing to do for webhooks is bearer tokens, IMO.  For SSE I'll have to think about it, but afaik you can bootstrap off the authorization needed on the subscription affordance.

benfrancis:
Following up from the Profile call today. If this is intended to refer to HTTP Basic Auth and RFC7235 _only_ allows credentials to provided in an `Authorization` header then my comment above is a moot point. HTTP Basic Auth can't be used by the SSE and WebSocket client-side JavaScript APIs in web browsers.

benfrancis:
In the long term this should probably just be fixed in the TD specification, or an HTTP binding document. See https://github.com/w3c/wot-thing-description/issues/1781

mlagally (approved):
The discussion was resolved and clarification was provided.

mlagally:
Profile call on March 232nd: Reviewed and approved to merge.

PR #381, opened 2023-03-24
Adding missing member "alternate" to the table

### LINKED ISSUE #336: Add link `alternate` relation type to table of link relation types
Assertion `common-constraints-links-media-types-5` says Consumers should interpret an `alternate` link relation type in a certain way, but this value is missing from the table in [section 6.6.1](https://w3c.github.io/wot-profile/#sec-link-relation-types).

mlagally:
Resolves #336

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/381.html" title="Last updated on Mar 14, 2024, 9:27 AM UTC (2589f13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/381/4af3498...2589f13.html" title="Last updated on Mar 14, 2024, 9:27 AM UTC (2589f13)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -802,6 +802,11 @@
               <td>link target is a collection member of the thing or thing model.</td>
               <td></td>
             </tr>
+            <tr>
+              <td><code>alternate</code></td>
+              <td>link target is a HTML based user interface of the thing.</td>
+              <td></td>
+            </tr>
 
           </tbody>
         </table>
--- a/index.html
+++ b/index.html
@@ -804,7 +804,7 @@
             </tr>
             <tr>
               <td><code>alternate</code></td>
-              <td>link target is a HTML based user interface of the thing.</td>
+              <td>link target is an alternative representation of the Thing.</td>
               <td></td>
             </tr>
 


index.html
```diff
@@ -802,6 +802,11 @@ <h3>Link Relation Types</h3>
               <td>link target is a collection member of the thing or thing model.</td>
               <td></td>
             </tr>
+            <tr>
+              <td><code>alternate</code></td>
+              <td>link target is a HTML based user interface of the thing.</td>
```
benfrancis, line 807:
```suggestion
              <td>link target is an alternative representation of the Thing.</td>
```

link rel=alternative could also be used for other use cases, such as linking to an XML version of the Thing Description for example. It's only when the type is text/html that it SHOULD be interpreted as a user interface.

benfrancis (changes requested):

index.html
egekorkan: It can also be a documentation website.

benfrancis: @egekorkan That's what the `service-doc` link relation is used for. There's an assertion which specifically says that link rel=alternative with type=text/html should be interpreted as a user interface, but a different content type could be used to mean something else.

lu-zero (approved):
It looks fine to me.

benfrancis (approved):

Issue #375, opened 2023-03-22
Review of HTTP Webhook Profile


benfrancis:
I have been evaluating the latest draft of the Webhook Profile with a view to potential implementation in WebThings Gateway (Producer) and WebThings Cloud (Consumer) and I think it needs some more work before it's in an implementable state. I've provided my feedback section by section below, with action items as task list items which can be turned into separate issues if necessary.

# 10 HTTP Webhook Profile
> This section defines the HTTP Webhook Profile, including a [Protocol Binding](https://w3c.github.io/wot-profile/#http-webhook-profile-protocol-binding) for observing properties and listening for events using [WebHooks](https://w3c.github.io/wot-profile/#sec-http-webhook-profile).

So far there's only a protocol binding for events.
- [x] Define a protocol binding for `observeproperty`, `unobserveproperty`, `observeallproperties` and `unobserveallproperties` (#376).

# 10.1 Introduction

Suggested refinements:
- [x] "Webhooks are a mechanism for Consumers to subscribe to events emitted by a Web Thing." -> "Webhooks can be used as a mechanism for Consumers to subscribe to events and observe properties of a Web Thing."
- [x] "Consumers implement a Webhook listener that is handling the event streams generated by Things." -> "Consumers implement a Webhook listener which handles events emitted by Things." (Webhooks do not use streams, they emit each event in a separate HTTP request which requires its own TCP connection. Some payload formats allow for multiple events to be batched into a one payload, but that is not "streaming".)
- [x] "The client behavior is used for the initial subscription operation, whereas the server behavior is used to accept the event streams sent by the Thing." -> "The client behavior is used for the initial subscription operation, whereas the server behavior is used to accept the events sent by the Thing."

### 10.1.1 Webhook Example

I think this example could use work. I'm not sure why the fireAlarm event would be fired when the button has been re-armed, and I'm not sure how a Consumer would know anything about sprinkler status. The cancellation payload format and unsubscribe protocol binding need updating to match the protocol binding in the specification text, i.e. `DELETE` instead of `POST` and provide the subscription ID in the URL rather than the body (a DELETE request does not usually contain a body).

- [ ] #389

#### 10.4.1.1 `subscribeevent`

- [x] Add `Content-Type` header to list of features of the subscription request, as in the example

> This involves the Web Thing initially sending an HTTP response to the Consumer with:
> - Status code set to 200
> - Content-Type header set to application/json
> - Accept header set to application/json
> 
> EXAMPLE 46
> ```
[...]

PR #382, opened 2023-03-24
webhook properties section

### LINKED ISSUE #376: Specify observe property protocol bindings for HTTP Webhook Profile
Currently the HTTP Webhook Profile specifies protocol bindings for subscribing and unsubscribing from events, but not yet for observing and unobserving properties. It says in the introduction that it covers both, so protocol bindings need to be defined for the following operations:

- `observeproperty`
- `unobserveproperty`
- `observeallproperties`
- `unobserveallproperties`

mlagally:
resolves #376

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/382.html" title="Last updated on Apr 19, 2023, 11:31 AM UTC (09ce906)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/382/27d0390...09ce906.html" title="Last updated on Apr 19, 2023, 11:31 AM UTC (09ce906)">Diff</a>


mlagally:
Ready for review.

index.html
```diff
@@ -2968,12 +2969,305 @@ <h2>Protocol Binding</h2>
...
+          </ul>
+          <pre class="example">
+              HTTP/1.1 200 OK
+              Content-Type: application/json
+              {
+                  subscriptionId: 1234-4544-1211
+              }
+          </pre>
+          <p><span class="rfc2119-assertion" id="http-webhook-profile-protocol-binding-observeproperty-6">
+              Whenever an event of the specified type occurs, the Web Thing MUST
```
egekorkan, line 3077: what is meant by event of the specified type?

Also the assertion id seems to indicate property (mentioning observeproperty) but text says events. They are sometimes similar but the semantics behind are different.

index.html
```diff
@@ -2968,12 +2969,305 @@ <h2>Protocol Binding</h2>
...
+  
+          <section class="ednote"> This is inconsistent: there are currently two different unobserve mechanisms
+            - subscriptionId in URL or in payload
+          </section>
+  
+          <p>
+            <span class="rfc2119-assertion" id="http-webhook-profile-protocol-binding-unobserveproperty-1">
+              In order to stop observing a property, a Consumer MUST terminate
+              the corresponding subscription with the Web Thing using the
+              <code>unobserve</code> operation of the <code>Property</code> resource.</span>
```
egekorkan, line 3095:
```suggestion
              <code>unobserveproperty</code> operation of the <code>Property</code> resource.</span>
```

egekorkan (changes requested):
Mostly minor operation names

index.html
mlagally: Thanks for noticing - That's a copy paste error - fixed

benfrancis:
Many of the points from my review of the events protocol binding (#375) also apply to this proposed properties binding, since the assertions have been copied across from there. I would suggest resolving those points before landing this and duplicating all of the same issues.

In particular, I would recommend:
- [ ] Change `observeproperty` and `observeallproperties` responses to `201` with a subscription URI in a header rather than an ID in the body. Add a `Content-type` header and remove the `Accept` header on `observeproperty` and `observeallproperties` requests to reflect this.
- [ ] Change `unobserveproperty` and `unobserveallproperties` requests to a `DELETE` on the subscription URL, with a `204` response and an example

I think these changes are important so that the protocol binding doesn't rely on a fixed URL structure (which we have so far avoided in all of the other protocol bindings). This would also resolve the current issue with two different unsubscribe mechanisms, one of which has strange semantics (`POST` to `/properties/level` is subscribe and `POST` to `properties/level/123-456-789` is unsubscribe), and the other is ambiguous (both operations are `POST` requests on the same URL, so presumably would have to be differentiated by the content in the body though that isn't shown). From a REST point of view it makes more sense to me to create a subscription resource with the ID in its URI to represent a subscription, then delete that resource to cancel the subscription. This would be consistent with what we do in the actions protocol binding for cancelling something.

The subscription URI could alternatively be included in both a header _and_ the body of the `observeproperty` response (which was the compromise we came up with for the actions protocol binding), but it should definitely be a URI and not just an ID.

Much of my feedback on the Event Connections section in #375 also applies here:
- [ ] Remove the assertion "If the connection between the Web Thing and the Consumer drops, the Web Thing MUST re-establish the connection."
- [ ] Remove the assertion "Once the connection is re-established the Web Thing SHOULD, if possible, send any missed notifications which occurred since the last successful notification." 

(These assertions have been copied and pasted from the SSE protocol binding, but they don't make sense for Webhooks where every property reading would be sent in a separate HTTP request, and therefore a separate TCP "connection".)

- [ ] Add assertions to say that property readings must be sent from the Thing to the Consumer with an HTTP `POST` request with a `Content-Type` header set to `application/json`, with a `200` or `204` response.

My suggestions about batched event payloads, event rate limiting, mutual authentication and abuse protection also apply here, but can be addressed separately.

index.html
```diff
@@ -3330,41 +3330,34 @@ <h4><code>subscribeevent</code></h4>
...
             <li>
-              <code>Accept</code> header set to <code>application/json</code>
+              <code>Location</code> header set to the URL of the <code>Event</code> resource
+              followed by a unique <code>subscriptionId</code>
             </li>
           </ul>
           <pre class="example">
-              HTTP/1.1 200 OK
+              HTTP/1.1 201 Created
               Content-Type: application/json
```
lu-zero, line 3345: is the Content-Type still needed?

mlagally:
@benfrancis 
thanks for your thorough review. I addressed most points, also see corresponding PRs #390 #391 and #392

index.html
mlagally:
```suggestion
```

mlagally:
Approve to merge in Profile call on April 19th

PR #393, opened 2023-04-26
Re-generate template.csv for 2023-04 TestFest

benfrancis:
This is a new template.csv for the [2023-04 TestFest](https://github.com/w3c/wot-testing/tree/main/events/2023.03.Online) (delayed from 2023-03).

This was generated using [extract.js](https://github.com/w3c/wot-profile/blob/main/testing/extract.js) which takes the latest specification from https://raw.githubusercontent.com/w3c/wot-profile/main/index.html and generates a list of assertions from it. I manually removed the comments it generates about the assertion tester.

There are different scripts and different output formats to choose from so I'm not sure which format you prefer, but it would be good to have an updated list of assertions for testing since they have changed quite a lot since the last TestFest.


mmccool:
Can merge, but I'll probably update again when I regenerate the IR after merging your test input.  The important thing about the format is the column titles.  My tooling will ignore the sequence number in particular, so it doesn't matter whether or not it is in there.

PR #408, opened 2024-04-30
Make sychronous member of ActionAffordance mandatory - closes #405

### LINKED ISSUE #405: [HTTP Basic Profile] Require that async actions set the synchronous member of an ActionAffordance to false
Currently the HTTP Basic Profile says:

> If the synchronous member of the [ActionAffordance](https://w3c.github.io/wot-thing-description/#actionaffordance) [[wot-thing-description11](https://w3c.github.io/wot-profile/#bib-wot-thing-description11)] is undefined then the Web Thing MAY respond with either a [Synchronous Action Response](https://w3c.github.io/wot-profile/#sync-action-response) or [Asynchronous Action Response](https://w3c.github.io/wot-profile/#async-action-response).

This might cause a generic WoT Consumer which doesn't implement the profile to get confused if it invokes an action, expects a synchronous response containing the action output, but actually gets an asychronous response with an ActionStatus payload linking to a dynamic resource.

We could consider asserting that if a Thing implements an asynchronous action then it must set the synchronous member of the ActionAffordance to false, to be more explicit that it will not respond synchronously.

This should then be enough to tell a generic Consumer not to expect the action output in the response to the invokeaction request, even if it can't then consume the queryaction and cancelaction operations.

Currently it's technically possible that the same action affordance to respond either synchronously or asynchronously depending on the anticipated duration of the action, and for a sufficiently sophisticated Consumer to understand both responses. This change would remove that flexibility in favour of more predictable behaviour, and a "progressive enhancement" approach. Consumers which don't support asynchronous actions can either "fire and forget" an action invocation, or ignore an asynchronous ActionAffordance altogether.

benfrancis:
closes #405

I don't like that this adds an additional mandatory term to ActionAffordances, but if this really is currently the only case where a Thing compliant with the HTTP Basic Profile will break a generic Consumer, then I think it's worth being more strict.

(I would also be happy to just make it mandatory for async actions to set `synchronous` to `false`)

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/408.html" title="Last updated on Apr 30, 2024, 2:58 PM UTC (cf660f7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/408/9a996de...benfrancis:cf660f7.html" title="Last updated on Apr 30, 2024, 2:58 PM UTC (cf660f7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -969,6 +969,7 @@
               "fade": {
                 "title": "Fade",
                 "description": "Fade the lamp to a given level",
+                "synchronous": false,
                 "input": {
                   "type": "object",
                   "properties": {
@@ -1366,6 +1367,14 @@
           </div>
           <p>
             <span class="rfc2119-assertion"
+            id="http-basic-profile-protocol-binding-invokeaction-24">
+              The <code>synchronous</code> member of an 
+              <a href="https://w3c.github.io/wot-thing-description/#actionaffordance"><code>ActionAffordance</code></a>
+              MUST be set to <code>true</code> or <code>false</code>.
+            </span>
+          </p>
+          <p>
+            <span class="rfc2119-assertion"
             id="http-basic-profile-protocol-binding-invokeaction-7">
               If the <code>synchronous</code> member of the
               <a href="https://w3c.github.io/wot-thing-description/#actionaffordance"><code>ActionAffordance</code></a>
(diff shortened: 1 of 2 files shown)


lu-zero (approved):
It is a step in the right direction, agreed also in the TF meeting.

PR #409, opened 2024-05-02
Update readme with the link to mastodon

lu-zero:
As discussed in https://github.com/w3c/wot-marketing/issues/461


lu-zero:
As discussed in the TF meeting, this can land, we can have a new readme according to the new marketing template later.

PR #415, opened 2024-06-18
Out of the box clarification

lu-zero:
To clarify the out-of-the-box section I moved up the terminology section since few terms from there are used there and I added the, hopefully, shortest explanation of it.

***
<a href="https://pr-preview.s3.amazonaws.com/luminem/wot-profile/pull/415.html" title="Last updated on Jun 19, 2024, 6:48 PM UTC (52e6d93)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/415/f68e08a...luminem:52e6d93.html" title="Last updated on Jun 19, 2024, 6:48 PM UTC (52e6d93)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -267,7 +267,11 @@
     <section id="out-of-the-box-interoperability">
       <h2>Out-of-the-Box Interoperability</h2>
       <p>
-        The definition of out-of-the-box interoperability used in this
+        At a high level, out-of-the-box interoperability means that a <a href="#dfn-consumer">Consumer</a> is guaranteed to be able to
+        use every capability of a <a href="#dfn-thing">Thing</a>, without Thing-specific customization.
+      </p>
+      <p>
+        The full definition of out-of-the-box interoperability used in this
         specification includes multiple layers.
         The following classification adopts terminology from the
         "H2020 – CREATE-IoT Project - Recommendations for commonalities and
--- a/index.html
+++ b/index.html
@@ -122,14 +122,14 @@
       This specification defines a <a href="#profiling-mechanism">Profiling
       Mechanism</a> and a set of <a href="#dfn-profile">Profiles</a>
       which enable <a href="#out-of-the-box-interoperability">out-of-the-box
-      interoperability</a> between 
+      interoperability</a> between
       <a href="#dfn-thing">Web Things</a> and their
-      <a href="#dfn-consumer">Consumers</a> on the 
+      <a href="#dfn-consumer">Consumers</a> on the
       <a href="https://www.w3.org/wot">Web of Things</a>.
     </p>
     <p>
       Being out-of-the-box interoperable means that any Consumer which conforms
-      with a given Profile can interact with any Thing which conforms with the 
+      with a given Profile can interact with any Thing which conforms with the
       same profile, without additional customization.
     </p>
     <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -243,32 +292,36 @@ <h2>Motivation</h2>
...
-        for Profiles were derived from these use cases. 
+        for Profiles were derived from these use cases.
       </p>
     </section>
 
     <section id="out-of-the-box-interoperability">
       <h2>Out-of-the-Box Interoperability</h2>
+      <p>
+        In layman terms out-of-box-interoperability implies that every capability of a conformant
+        device can be perused by a conformant client, within the scope of the profile.
```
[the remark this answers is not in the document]
egekorkan: Also I am not sure peruse is the correct term here. I would just use "use" or activating a capability?

egekorkan (changes requested):
Also, I think that terminology is always after introduction in W3C documents. In TD and Arch, it is like that.

index.html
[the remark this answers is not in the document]
lu-zero: Yes, you are right. (folding and rebasing)

egekorkan:
> If we use the terms in parts of the introduction the terminology has to come first.

I agree. It is generally the case in all introductions but terminology still comes later, for some reason I do not know. My only guess is to have an introduction without making the user scroll a long time to get there, which can happen if the terminology is long.

index.html
```diff
@@ -243,32 +292,36 @@ <h2>Motivation</h2>
...
+        for Profiles were derived from these use cases.
       </p>
     </section>
 
     <section id="out-of-the-box-interoperability">
       <h2>Out-of-the-Box Interoperability</h2>
+      <p>
+        In layman terms, out-of-box-interoperability implies that every capability of a conformant
+        Thing can be used by a conformant Consumer, within the scope of the profile.
+      </p>
```
benfrancis, line 321: I'm not sure these are really layman's terms 😆, and the definitions of "profile" and "out-of-the-box interoperability" would depend on each other.

How about...

```suggestion
      <p>
        At a high level, out-of-the-box interoperability means that a Consumer is guaranteed to be able to
        use every capability of a Thing, without Thing-specific customization.
      </p>
```

index.html
```diff
@@ -243,32 +292,36 @@ <h2>Motivation</h2>
...
       <p>
         The definition of out-of-the-box interoperability used in this
-        specification includes multiple layers. 
+        specification includes multiple layers.
```
benfrancis, line 324: With the addition of the shorter high level definition above, we might want to make this say "The _full_ definition of out-of-the-box interoperability used in this specification includes multiple layers."

index.html
```diff
@@ -171,69 +171,118 @@
...
     <p>
-      Future versions of this specification, or extension specifications, may 
+      Future versions of this specification, or extension specifications, may
       define additional Profiles.
     </p>
   </section>
 
   <section id='sotd'></section>
 
+  <section id="terminology">
```
benfrancis, line 181: We should probably leave this where it is for consistency with the other WoT specifications, but could link the word "profile" in the introduction to its definition in the terminology section for clarity.

benfrancis (changes requested):

index.html
lu-zero: I opened the discussion about it during today's main call. Cross linking is indeed a better solution.

lu-zero:
Approved as per today TF call.

PR #417, opened 2024-07-17
A start on use cases and requirements for Profiles 2.0

benfrancis:
Following on from my [strawman proposal](https://lists.w3.org/Archives/Public/public-wot-wg/2024Feb/0000.html) for Profiles 2.0, I think it's important that this time around we agree on a clear set of use cases and requirements up-front before we start work on the specification.

With this in mind I have made a start on a Use Cases & Requirements document which so far includes:
- Introduction - A preamble including recent text from the existing 1.0 specification which describes the general motivation for profiles
- Use Cases - A start on a collection of generic and domain-specific use cases based on my own direct experience, plus an initial set of headings for other potential application domains as an invitation for others to add their own use cases
- Requirements - An initial set of proposed requirements for a Profiles 2.0 specification (which would just define the profiling mechanism and a registry of profiles) and for individual profile documents

I welcome your feedback.

### DIFF
--- a/requirements/V2_REQUIREMENTS.md
+++ b/requirements/V2_REQUIREMENTS.md
@@ -1,0 +1,69 @@
+# WoT Profiles 2.0 Use Cases & Requirements
+
+## Introduction
+
+The [Web of Things](https://www.w3.org/WoT/) (WoT) seeks to counter the fragmentation of the [Internet of Things](https://en.wikipedia.org/wiki/Internet_of_things) (IoT) by using and extending existing, standardized web technologies.
+
+The [W3C WoT Thing Description 1.1 specification](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) defines an information model and JSON-based representation format for describing the capabilities of connected devices and the interfaces with which to communicate with them. Thing Descriptions are designed to be protocol-agnostic and flexible enough to describe a wide range of existing ("brownfield") IoT devices.
+
+In order to provide this level of flexibility the Thing Description specification includes a number of extension points including protocol bindings, payload bindings, security mechanisms, link relations and semantic contexts. As long as all of the capabilities of a device can be described using a Thing Description and a [Consumer](https://w3c.github.io/wot-profile/#dfn-consumer) implements all of the extensions used, the Consumer should be able to interoperate with that device. However, the result of this extensible architecture is that any given Consumer can only interoperate with a subset of possible [Web Things](https://w3c.github.io/wot-profile/#dfn-thing).
+
+The WoT Profiles specification should complement the [Thing Description](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) specification, by enabling ad-hoc interoperability through the use of "profiles". A profile prescribes a finite set of extensions and defaults that a Thing can be constrained to in order to guarantee out-of-the-box interoperability with any Consumer which implements that profile. Out-of-the-box interoperability means that a [Consumer](https://w3c.github.io/wot-profile/#dfn-consumer) is guaranteed to be able to use every capability of a [Thing](https://w3c.github.io/wot-profile/#dfn-thing), without Thing-specific customization.
+
+Profiles will be designed specifically for new ("greenfield") implementations where developers have the freedom to conform to a prescriptive protocol binding and set of common constraints, in order to benefit from this additional level of interoperability.
+
+## Use Cases
+
+As a developer of a WoT Consumer, I want to guarantee out-of-the-box interoperability with a wide range of WoT Things.
+
+As a developer of a WoT Thing, I want to guarantee out-of-the-box interoperability with a wide range of WoT Consumers.
+
+### Smart Home
+
+As the developer of DIY smart home hub software, I want to safely expose an HTTP-based API for monitoring and controlling a home which conforms to a common standard, so that cloud services from multiple vendors can add value to a user's smart home without having to implement a vendor-specific API.
+
+As the developer of a smart home mobile app I want to offer my users out-of-the-box interoperability with a wide range of smart home devices so that I can maximise the audience for my app.
+
+### Smart Buildings
+
+As the developer of a smart building hub, I want to safely expose an HTTP-based API for monitoring and controlling a commercial building which conforms to a common standard, so that cloud services from multiple vendors can consume data from that hub without having to implement a vendor-specific API.
+
+As the developer of a smart building cloud service, I want to be able to provide data analytics for smart buildings using smart building hubs from different vendors which conform to a common standard, rather than having to implement multiple vendor-specific APIs.
+
+### Smart Cities
+
+### Manufacturing
+
+### Energy
+
+### Transportation
+
+### Agriculture
+
+## Requirements
+
+### WoT Profiles Specification
+
+The WoT Profiles specification:
+1. MUST define a mechanism by which the author of a WoT Thing Description can denote that the Thing it describes conforms to a particular profile
+2. MUST define a registry of profiles to which WoT Things and WoT Consumers may conform
+3. SHOULD aim to keep the number of registered profiles as small as possible by rejecting profiles that significantly overlap in technologies or use cases, in order to minimise fragmentation on the Web of Things
+### Individual Profiles
+
+Individual profile documents:
+1. MUST specify an identifier (URI) which identifies the profile 
+2. SHOULD constrain a finite set of protocol bindings that a conformant Thing may use and that a conformant Consumer can be expected to support
+	1. MAY constrain conformant Things to always follow certain defaults defined in a protocol binding document
+3. SHOULD constrain a finite set of payload bindings that a conformant Thing may use and that a conformant Consumer can be expected to support
+	1. MAY constrain conformant Things to always follow certain default payload formats defined in a payload binding document
+	2. MAY constrain conformant Things to follow certain conventions such as date formats and error formats
+4. SHOULD constrain a finite set of security mechanisms that a conformant Thing may use and that a conformant Consumer can be expected to support
+5. SHOULD constrain a finite set of discovery mechanisms that a conformant Thing may use and that a conformant Consumer can be expected to support
+6. MAY constrain a finite set of link relation types that a conformant Thing is recommended to use, and how they should be interpreted by a WoT Consumer
+7. MAY constrain a finite set of semantic contexts that a conformant Thing is recommended to use and that a conformant Consumer can be expected to support
+8. SHOULD NOT directly extend the WoT Thing Description information model, which should be defined using TD context extensions
+9. SHOULD NOT define protocol behaviour, which should be defined in a protocol binding document
+10. SHOULD NOT define data payload formats, which should be specified in a payload binding document
+11. SHOULD NOT define discovery mechanisms, which should be defined in the WoT Discovery specification
+12. SHOULD NOT define security mechanisms, which should be defined in a protocol binding document
+13. SHOULD NOT describe an existing single-vendor IoT platform, which should be defined in a platform binding document


lu-zero:
The wot-uc group has [this form](https://github.com/w3c/wot-usecases/issues/new?assignees=&labels=UC&projects=&template=WoT-UseCases-Template.yml&title=Put+your+title+for+the+use+case.) for submitting new use-cases, we planned to give it a test with the TD taskforce, but we could try also for profiles.

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,69 @@
...
+
+The WoT Profiles specification should complement the [Thing Description](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) specification, by enabling ad-hoc interoperability through the use of "profiles". A profile prescribes a finite set of extensions and defaults that a Thing can be constrained to in order to guarantee out-of-the-box interoperability with any Consumer which implements that profile. Out-of-the-box interoperability means that a [Consumer](https://w3c.github.io/wot-profile/#dfn-consumer) is guaranteed to be able to use every capability of a [Thing](https://w3c.github.io/wot-profile/#dfn-thing), without Thing-specific customization.
```
TallTed, line 11: This might want further massage to be clear. Without something like this tweak, the last sentence suggests that a Consumer which has implemented *any* profile can use every capability of *any* Thing, not limited to Things that conform to the profile(s) implemented by the Consumer.
```suggestion
The WoT Profiles specification should complement the [Thing Description](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) specification, by enabling ad-hoc interoperability through the use of "profiles". A profile prescribes a finite set of extensions and defaults that members of a class of Thing can be constrained to, in order to guarantee out-of-the-box interoperability with any Consumer which implements that profile. Out-of-the-box interoperability means that a [Consumer](https://w3c.github.io/wot-profile/#dfn-consumer) is guaranteed to be able to use every capability of a member of a class of [Thing](https://w3c.github.io/wot-profile/#dfn-thing), without Thing-specific customization.
```

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,69 @@
...
+
+As a developer of a WoT Consumer, I want to guarantee out-of-the-box interoperability with a wide range of WoT Things.
+
+As a developer of a WoT Thing, I want to guarantee out-of-the-box interoperability with a wide range of WoT Consumers.
+
+### Smart Home
+
+As the developer of DIY smart home hub software, I want to safely expose an HTTP-based API for monitoring and controlling a home which conforms to a common standard, so that cloud services from multiple vendors can add value to a user's smart home without having to implement a vendor-specific API.
+
+As the developer of a smart home mobile app I want to offer my users out-of-the-box interoperability with a wide range of smart home devices so that I can maximise the audience for my app.
```
TallTed, line 25:
```suggestion
As the developer of a smart home mobile app, I want to offer my users out-of-the-box interoperability with a wide range of smart home devices so that I can maximise the audience for my app.

As the developer of a smart home device, I want my device to be controllable by any smart home hub software and/or any smart home mobile app that conforms to a common standard....
```

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Requirements
+
+### WoT Profiles Specification
+
+The WoT Profiles specification:
+1. MUST define a mechanism by which the author of a WoT Thing Description can denote that the Thing it describes conforms to a particular profile
+2. MUST define a registry of profiles to which WoT Things and WoT Consumers may conform
+3. SHOULD aim to keep the number of registered profiles as small as possible by rejecting profiles that significantly overlap in technologies or use cases, in order to minimise fragmentation on the Web of Things
+### Individual Profiles
```
TallTed, line 51: These are not dependent nor subsequent requirements; each stands on its own. Therefore, this should be an unordered list (bullets) not ordered (numbers). 

If there is a specific need to refer back to individual requirements, each requirement in these two lists (including the sub-requirements) SHOULD be given its own permanent identifier (URI), which can be used in future documents here and elsewhere.
```suggestion
* MUST define a mechanism by which the author of a WoT Thing Description can denote that the Thing it describes conforms to a particular profile
* MUST define a registry of profiles to which WoT Things and WoT Consumers may conform
* SHOULD aim to keep the number of registered profiles as small as possible by rejecting profiles that significantly overlap in technologies or use cases, in order to minimise fragmentation on the Web of Things

### Individual Profiles
```

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,69 @@
...
+5. SHOULD constrain a finite set of discovery mechanisms that a conformant Thing may use and that a conformant Consumer can be expected to support
+6. MAY constrain a finite set of link relation types that a conformant Thing is recommended to use, and how they should be interpreted by a WoT Consumer
+7. MAY constrain a finite set of semantic contexts that a conformant Thing is recommended to use and that a conformant Consumer can be expected to support
+8. SHOULD NOT directly extend the WoT Thing Description information model, which should be defined using TD context extensions
+9. SHOULD NOT define protocol behaviour, which should be defined in a protocol binding document
+10. SHOULD NOT define data payload formats, which should be specified in a payload binding document
+11. SHOULD NOT define discovery mechanisms, which should be defined in the WoT Discovery specification
+12. SHOULD NOT define security mechanisms, which should be defined in a protocol binding document
+13. SHOULD NOT describe an existing single-vendor IoT platform, which should be defined in a platform binding document
```
TallTed, line 69:
```suggestion
* MUST specify an identifier (URI) which identifies the profile 
* SHOULD constrain a finite set of protocol bindings that a conformant Thing may use and that a conformant Consumer can be expected to support
  * MAY constrain conformant Things to always follow certain defaults defined in a protocol binding document
* SHOULD constrain a finite set of payload bindings that a conformant Thing may use and that a conformant Consumer can be expected to support
  * MAY constrain conformant Things to always follow certain default payload formats defined in a payload binding document
  * MAY constrain conformant Things to follow certain conventions such as date formats and error formats
* SHOULD constrain a finite set of security mechanisms that a conformant Thing may use and that a conformant Consumer can be expected to support
* SHOULD constrain a finite set of discovery mechanisms that a conformant Thing may use and that a conformant Consumer can be expected to support
* MAY constrain a finite set of link relation types that a conformant Thing is recommended to use, and how they should be interpreted by a WoT Consumer
* MAY constrain a finite set of semantic contexts that a conformant Thing is recommended to use and that a conformant Consumer can be expected to support
* SHOULD NOT directly extend the WoT Thing Description information model, which should be defined using TD context extensions
* SHOULD NOT define protocol behaviour, which should be defined in a protocol binding document
* SHOULD NOT define data payload formats, which should be specified in a payload binding document
* SHOULD NOT define discovery mechanisms, which should be defined in the WoT Discovery specification
* SHOULD NOT define security mechanisms, which should be defined in a protocol binding document
* SHOULD NOT describe an existing single-vendor IoT platform, which should be defined in a platform binding document
```

TallTed (changes requested):

requirements/V2_REQUIREMENTS.md
benfrancis: @TallTed Thanks for this feedback. The out-of-the-box interoperability definition was taken out of context from the current specification, I've tweaked the text and it's hopefully now more clear.

requirements/V2_REQUIREMENTS.md
benfrancis: I started the Use Cases & Requirements document as a Markdown file for simplicity, and as far as I know GitHub Markdown doesn't support anchors in list items. A numbered list seems like a reasonable way of making individual requirements referenceable.

If someone wants to turn this into a more formal HTML document using ReSpec (which supports such things) then feel free, but I've [done this before](https://w3c.github.io/web-thing-protocol/requirements/) and for this kind of document I tend to think it just makes the document harder to edit.

benfrancis:
@lu-zero wrote:
> The wot-uc group has [this form](https://github.com/w3c/wot-usecases/issues/new?assignees=&labels=UC&projects=&template=WoT-UseCases-Template.yml&title=Put+your+title+for+the+use+case.) for submitting new use-cases, we planned to give it a test with the TD taskforce, but we could try also for profiles.

That seems very overcomplicated for our needs, and I would be concerned that all the mandatory fields will deter people from contributing.

Perhaps we could compromise on something structured but simpler, like a user story:

As a _______ I want to _____ so that ________

Application domain: [Smart Home|Smart Buildings|Smart Cities|Manufacturing|Energy|Transportation|Agriculture|Other]

---

Rather than formatting and grammar, I'm much more interested in feedback on the actual content. Does this match people's understanding of what profiles are and what they are used for?

I'm keen to assess the level of consensus around that because I'm concerned from the past few months of discussions that there's still a lack of a basic shared understanding.

danielpeintner (commented):
My point of view ...

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,71 @@
...
+
+The WoT Profiles specification should complement the [Thing Description](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) specification, by enabling ad-hoc interoperability through the use of "profiles". A profile prescribes a finite set of extensions and defaults that a Thing can be constrained to in order to guarantee out-of-the-box interoperability with any Consumer which implements that profile. Out-of-the-box interoperability means that a [Consumer](https://w3c.github.io/wot-profile/#dfn-consumer) which implements a given profile is guaranteed to be able to use every capability of a [Thing](https://w3c.github.io/wot-profile/#dfn-thing), which implements that same profile, without Thing-specific customization.
```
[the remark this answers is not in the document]
benfrancis: @danielpeintner But this is exactly what it says. In particular, see points 8-13, one of which says that profiles "SHOULD NOT directly extend the WoT Thing Description information model".

When I say that Profiles are meant to complement the Thing Description specification, I just mean that they should not conflict with it. All of the requirements listed for profiles are about _constraining_ extension points, not about extending anything. However, describing a profile as being a "subset of the TD" is also not an accurate description, because it constrains much more than than just the Thing Description.

I feel you may be reacting to preconceived notions of what you think profiles are, or what has been discussed in the past, rather than the new approach described here and in the [strawman proposal](https://lists.w3.org/Archives/Public/public-wot-wg/2024Feb/0000.html).

Is there a way that this could be made clearer?

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,69 @@
...
+
+The [W3C WoT Thing Description 1.1 specification](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) defines an information model and JSON-based representation format for describing the capabilities of connected devices and the interfaces with which to communicate with them. Thing Descriptions are designed to be protocol-agnostic and flexible enough to describe a wide range of existing ("brownfield") IoT devices.
+
+In order to provide this level of flexibility the Thing Description specification includes a number of extension points including protocol bindings, payload bindings, security mechanisms, link relations and semantic contexts. As long as all of the capabilities of a device can be described using a Thing Description and a [Consumer](https://w3c.github.io/wot-profile/#dfn-consumer) implements all of the extensions used, the Consumer should be able to interoperate with that device. However, the result of this extensible architecture is that any given Consumer can only interoperate with a subset of possible [Web Things](https://w3c.github.io/wot-profile/#dfn-thing).
```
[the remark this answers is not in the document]
TallTed: To a degree. In my experience, using it eliminates possible (and never adds) confusion, while leaving it out often (though not always) leaves possible confusion, so I almost always recommend it in spec-type documents.

requirements/V2_REQUIREMENTS.md
TallTed: > GitHub Markdown doesn't support anchors in list items. 

Drat. Markdown constantly surprises me with features it lacks, especially its site-to-site variance.

> A numbered list seems like a reasonable way of making individual requirements referenceable.

It's reasonable *if* it's reasonably-well guaranteed that the members of such lists will never change order or meaning, or at least that substitute "blanks" that say something like "this item has been deleted in favor of..." will be inserted to ensure that external references will always reach the appropriate item *or* such a "blank". Even then, it would be better to have number+short-title referenceability.

All things considered, including the [W3C Priority of Constituencies](https://www.w3.org/TR/design-principles/#priority-of-constituencies), I would strongly advise conversion to [ReSpec](https://respec.org/docs/) or [Bikeshed](https://speced.github.io/bikeshed/), with an appropriate anchor for each requirement.

requirements/V2_REQUIREMENTS.md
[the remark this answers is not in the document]
danielpeintner: @benfrancis Very good if we share the same understanding.
I tried to read it objectively. However, maybe I misunderstood some statements (I'm not a native speaker).

I have taken the liberty of suggesting some changes that would make it clearer to me.

```suggestion
The WoT Profiles specification should comply with the [Thing Description](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) specification, by enabling ad-hoc interoperability through the use of "profiles". A profile prescribes defaults that a Thing can be constrained to in order to guarantee out-of-the-box interoperability with any Consumer which implements that profile. Out-of-the-box interoperability means that a [Consumer](https://w3c.github.io/wot-profile/#dfn-consumer) which implements a given profile is guaranteed to be able to use every capability of a [Thing](https://w3c.github.io/wot-profile/#dfn-thing), which implements that same profile, without Thing-specific customization.
```

On purpose I removed "finite set of extensions" since this means to me the opposite of what it should do. What do you see requiring being an extension?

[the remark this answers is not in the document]
benfrancis: @danielpeintner OK, I think I understand what you are trying to clarify, but these edits don't quite make sense to me so let's try to refine them.

> The WoT Profiles specification should comply with the [Thing Description](https://w3c.github.io/wot-profile/#bib-wot-thing-description11) specification, by enabling ad-hoc interoperability through the use of "profiles".

It doesn't seem right to say that the Profiles specification "complies" with the Thing Description specification. It would be correct to say that the Thing Descriptions of Web Things which implement a Profile should comply with the Thing Description specification, but I had thought that was implied (and is now already the case with Profiles 1.0). Perhaps it needs to be made more explicit.

Also, the Profile specification does not enable ad-hoc interoperability simply by complying with the Thing Description specification, so this sentence doesn't really make sense.

> A profile prescribes ~a finite set of extensions and~ defaults that a Thing can be constrained to in order to guarantee out-of-the-box interoperability with any Consumer which implements that profile.

A profile doesn't just prescribe defaults, it constrains a Web Thing to using a particular set of extensions.

But when I say that a profile prescribes a finite set of extensions, I don't mean that it _defines_ those extensions (as Profiles 1.0 arguably does), it just lists them (see the [HTTP Basic Profile 2.0](https://docs.google.com/document/d/1LjBWiqQZXi85gXP2NNckQni5os6dwxW6UDGrZDb3cns/edit#heading=h.71drcnjjomtn) strawman proposal for an example . The "extensions points" are defined in the previous paragraph as being:
- Protocol bindings
- Payload bindings
- Security mechanisms
- Discovery mechanisms
- Link relations
- Semantic contexts

These are the existing extension points I have identified in the Web of Things which need to be constrained in order for a Consumer to guarantee interoperability. Essentially all a Profile should do in this proposal is to list the allowed extensions for each of these points, not define any new extensions.

--

To try to address your feedback I have:

benfrancis:
I've also added a requirement to try to make it clear that a Thing should be able to conform to multiple profiles.

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,74 @@
...
+
+
+"WoT Profiles" are a mechanism by which out-of-the-box interoperability between WoT [Consumers](https://w3c.github.io/wot-profile/#dfn-consumer) and [Things](https://w3c.github.io/wot-profile/#dfn-thing) can be guaranteed, by constraining a Thing to a finite list of options for each extension point, and requiring that it conforms to certain defaults. As long as a Consumer implements all of the extensions and defaults prescribed by a Profile, it should be guaranteed to be able to use all of the capabilities of a Thing which conform to that Profile, without Thing-specific customisation.
+
+Profiles will be designed specifically for new ("greenfield") implementations where developers have the freedom to conform to a prescriptive protocol binding and set of common constraints, in order to benefit from this additional level of interoperability.
```
lu-zero, line 14: I think this is the part that may see contention.

WoT Greenfield is just another ecosystem's Brownfield. In order of having established ecosystems converge/evolve to WoT more easily we should have more than ad-hoc compatibility for them.

requirements/V2_REQUIREMENTS.md
lu-zero: Bikeshed feels the easiest way forward IMHO.

requirements/V2_REQUIREMENTS.md
```diff
@@ -0,0 +1,74 @@
...
+7. MAY constrain a finite set of semantic contexts that a conformant Thing is recommended to use and that a conformant Consumer can be expected to support
+8. SHOULD NOT directly extend the WoT Thing Description information model, which should be defined using TD context extensions
+9. SHOULD NOT define protocol behaviour, which should be defined in a protocol binding document
+10. SHOULD NOT define data payload formats, which should be specified in a payload binding document
+11. SHOULD NOT define discovery mechanisms, which should be defined in the WoT Discovery specification
+12. SHOULD NOT define security mechanisms, which should be defined in a protocol binding document
+13. SHOULD NOT describe an existing single-vendor IoT platform, which should be defined in a platform binding document
+14. MUST NOT include any assertions which would require a Thing Description of a conformant Web Thing to be non-conformant with the Thing Description 2.0 specification
+15. SHOULD NOT include any assertions which conflict with another Profile, so that a Thing may conform with multiple Profiles
```
lu-zero, line 74: Do we want to allow a thing to describe something outside of the profile scope?

e.g. a thing may sport an http profile that covers only properties and all the properties then have an http form constrained by the profile, but it can also provide coap forms and event affordances in the same description. 

The consumer then may just ignore them or process them using the degraded consumption rules we should write down in the TD document.

requirements/V2_REQUIREMENTS.md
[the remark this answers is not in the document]
danielpeintner: Thank you for the changes. It helps a lot. 

I was wondering whether we can also make it even more clear that profiles are meant to _use_ extension points and are never allowed to create *new* extensions (unless they go the usual way with for example creating a new binding or so etc)

requirements/V2_REQUIREMENTS.md
benfrancis: I'm surprised by this, because until recently I thought this was a key differentiation that everyone agreed on.

Of course I agree that today's "greenfield" is tomorrow's "brownfield", but really the important point here is the distinction between the "descriptive" approach of protocol binding templates vs. the "prescriptive" approach of profiles.

In WoT 1.0 there are essentially two separate approaches to WoT - Binding templates (poorly named IMO) are best for describing existing systems where the communication interface can not be changed, whereas profiles provide a template (🤔) for new implementations which have the freedom to conform to a common standard in return for an increased level of interoperability.

In WoT 2.0 I'm proposing that bindings and profiles work better together, with profiles adding an optional layer of prescriptive constraints on top of bindings.

I don't see any benefit in using Profiles to describe existing IoT platforms, that's what platform bindings are for.

However, given there now seems to be some resistance to the greenfield/brownfield terminology I have reworded this paragraph to say:

> Whilst the Web of Things provides the freedom to describe a wide range of existing IoT systems using [WoT Binding Templates](https://www.w3.org/TR/wot-binding-templates/), Profiles provide an optional additional layer of common constraints to which new implementations can conform in order to benefit from an increased level of interoperability.

Is that any better?

requirements/V2_REQUIREMENTS.md
benfrancis: Yes, and there is a [note](https://w3c.github.io/wot-profile/#h-note-0) to this effect in the existing 1.0 specification.

I've added this note the end of the introduction.

requirements/V2_REQUIREMENTS.md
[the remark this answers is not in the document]
benfrancis: I think this is kind of already covered by points 8-12, but I've added some text to the introduction to reinforce this high level point.

> Profiles are designed to constrain, not extend, the Web of Things. They should only be used to constrain the set of options for existing extension points, never to extend the Web of Things directly.

requirements/V2_REQUIREMENTS.md
```diff
@@ -11,7 +11,9 @@ In order to provide this level of flexibility the Thing Description specificatio
...
 
-Profiles will be designed specifically for new ("greenfield") implementations where developers have the freedom to conform to a prescriptive protocol binding and set of common constraints, in order to benefit from this additional level of interoperability.
+Whilst the Web of Things provides the freedom to describe a wide range of existing IoT systems using [WoT Binding Templates](https://www.w3.org/TR/wot-binding-templates/), Profiles provide an optional additional layer of common constraints to which new implementations can conform in order to benefit from an increased level of interoperability.
+
+Conforming to a Profile does not prevent a Web Thing from describing additional capabilities and protocol bindings in their Thing Description beyond those described in the Profile, as long as they conform with all of the normative assertions of the Profile.
```
TallTed, line 18:
```suggestion
Conforming to a Profile does not prevent a Web Thing from describing additional capabilities or protocol bindings beyond those described in the Profile in their Thing Description, as long as they conform with all of the normative assertions of the Profile. 
```

PR #419, opened 2025-05-28
Explicity state contentType of application/json in Forms - closes #141

### LINKED ISSUE #141: how to handle multiple forms
Do we need to define a set of rules how multiple forms are to be handled by the consumer.

IMHO we need a defined algorithm to ensure interop.

Please see comment in:
https://github.com/w3c/wot-profile/pull/125#issuecomment-966589941

benfrancis:
This PR adds text which explicitly says that a Consumer should select Forms where the `contentType` member is set to `application/json` (once defaults have been applied).

This is intended to resolve an ambiguity where there may be multiple Forms which satisfy the existing criteria (e.g. one using XML and one using JSON).

I've also added a Note about `application/json` being wrapped in `text/event-stream` in the case of SSE. I'm not completely sure this is the right approach, but was an idea proposed by @relu91 in https://github.com/w3c/wot-binding-templates/issues/347.

Ideally in TD 2.0 there will be a normative algorithm for selecting a Form, but in the meantime this will help reduce ambiguity in Profiles.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/419.html" title="Last updated on May 28, 2025, 11:23 AM UTC (fd01363)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/419/4165bf1...benfrancis:fd01363.html" title="Last updated on May 28, 2025, 11:23 AM UTC (fd01363)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1032,13 +1032,17 @@
             <ul>
               <li>
                 After defaults have been applied, its <code>op</code> member
-                contains the value <code>readproperty</code>.
+                contains the value <code>readproperty</code>
               </li>
               <li>
                 After being resolved against a base URL where applicable, the
                 <a href="https://tools.ietf.org/html/rfc3986#section-3.1">URI
                   scheme</a> [[RFC3986]] of the value of its <code>href</code>
                 member is <code>http</code> or <code>https</code>
+              </li>
+              <li>
+                After defaults have been applied, the value of its
+                <code>contentType</code> member is <code>application/json</code>
               </li>
             </ul>
           </div>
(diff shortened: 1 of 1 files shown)


benfrancis:
Profile call 2025-05-28

- Sebastian: For Thing Descriptions in general we may not need to specify this because application/json is the default and Consumers should have their own algorithm for selecting a Form.
- Ben: Currently possible for a Consumer to select XML for example, depends how specific we want to be.
- Cris: Reasonable to tell Consumers to look for Forms that have application/json. Could just say about applying defaults once.

Will leave the PR open to allow for async reviews

danielpeintner (commented):
_Just a pass-by comment (I didn't look more closely)._

The sections define the 3 steps to select the right form (op, http(s) and contentType) but do not state what happen if there is no match. Maybe I just don't see such a statement, if there is none, I think it is worth mentioning what happens...

egekorkan (approved):

PR #428, opened 2025-07-09
Replace Accessibility Constraints with Accessibility Considerations section - closes #320

### LINKED ISSUE #320: Accesibility assertions are not specific enough
As I have mentioned in #316 and in the call, I am objecting the assertions added in #316 in their current state.

It is not clear where the title and description terms are required in section 6.1. If it is for Thing level, we already need title as required by the TD spec. If it is for all levels, then it would mean having these terms literally everywhere in the TD, including data schemas, which would make TDs explode. The assertions MUST be testable and these are currently not specific enough. Testability is not only relevant for implementation report generation but if I am to implement these features, the spec is simply not guiding me enough.

@mlagally has asked me to provide a PR but I cannot do that until I understand what is required. The intention of the PR #316 is there (make TDs accessible) but it is not clear why these terms are needed and the relationship between the intention. If we want to make TDs accesible, title, titles, description and descriptions have to be present literally everywhere.

benfrancis:
As discussed in #320.

Closes #320

### DIFF
--- a/index.html
+++ b/index.html
@@ -456,31 +456,6 @@
       The following sections are applicable for profiles defined by this document.
     </p>
 
-    <!-- Identifiers TODO -->
-
-    <!-- Accessibility -->
-
-    <section id="common-constraints-accessibilityy">
-      <h2 id="accessibility">Accessibility</h2>
-      <p>Authors of <a>Thing Descriptions</a> must ensure that the things described by them are accessible to
- 	users with disabilities.
-      </p>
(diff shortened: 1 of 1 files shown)


egekorkan (approved):

lu-zero (approved):

lu-zero (approved):
I was tempted to refer to suggest to refer Zsye subtag in case somebody really needs a title with emojis...

PR #429, opened 2025-07-09
Remove Default Language section - closes #373

### LINKED ISSUE #373: Mismatch with MUST clause and examples
In [section 8. Default Language](https://www.w3.org/TR/2023/WD-wot-profile-20230118/#sec-default-language),

> One Map contained in an @context Array MUST contain a name-value pair that defines the default language for the Thing Description, where the name is the Term @language and the value is a well-formed language tag as defined by [BCP47] (e.g., en, de-AT, gsw-CH, zh-Hans, zh-Hant-HK, sl-nedis).

1. This section has MUST clause to include default language in `@context` array, but examples in this spec (e.g. [example 1](https://www.w3.org/TR/2023/WD-wot-profile-20230118/#example-1), [example 2](https://www.w3.org/TR/2023/WD-wot-profile-20230118/#example-2), [example 4](https://www.w3.org/TR/2023/WD-wot-profile-20230118/#example-4)) do not have such.
2. In [6.3.2 Human-Readable Metadata in WoT Things Description](https://www.w3.org/TR/wot-thing-description/#titles-descriptions-serialization-json), handling of language is discussed in detail and fallback is also defined (like quoted below). This section seems not to be consisted with these.

> When metadata such as @direction is not present, TD Consumers SHOULD use[ first-strong detection](https://www.w3.org/TR/i18n-glossary/#dfn-first-strong-detection) as a fallback. For the MultiLanguage [Map](https://www.w3.org/TR/wot-thing-description/#dfn-map), TD Consumers MAY infer the [base direction](https://www.w3.org/TR/i18n-glossary/#def_base_direction) from the language tag of the individual strings.

Also, there should be some mention about which language should be applied for reply of property request, write property, or action in protocol bindings.

benfrancis:
As discussed in #373.

Closes #373.

### DIFF
--- a/index.html
+++ b/index.html
@@ -867,18 +867,6 @@
           MUST conform with the Problem Details format [[RFC7807]].</span>
       </p>
     </section>
-
-      <!-- Default Language -->
-    <section id="sec-default-language">
-      <h2>Default Language</h2>
-      <p><span class="rfc2119-assertion" id="common-constraints-default-language">
-          One Map contained in an <code>@context Array</code> MUST contain a name-value pair
-          that defines the default language for the Thing Description,
-          where the name is the Term <code>@language</code> and the value
-          is a well-formed language tag as defined by [BCP47]
-          (e.g., en, de-AT, gsw-CH, zh-Hans, zh-Hant-HK, sl-nedis).</span>
-        </span>
-    </section>
   </section>
 
 


egekorkan (approved):

danielpeintner (commented):
I might be wrong, but I think the following line needs to change too.

https://github.com/w3c/wot-profile/blob/637100f5e588e74e956dd20d7c61bce52cd9fc8d/index.html#L171

benfrancis:
@danielpeintner wrote:
> I might be wrong, but I think the following line needs to change too.
> 
> https://github.com/w3c/wot-profile/blob/637100f5e588e74e956dd20d7c61bce52cd9fc8d/index.html#L171

Yep, thanks. I've changed this too.

Issue #347, opened 2023-01-11
Clarify OAuth schemes - consider requiring Bearer tokens only


mlagally:
There are significant implementation efforts required to address the code and client flows for OAuth.
Bearer Tokens are the underlying mechanism - consider lowering the bar on that authentication requirement.

benfrancis:
Where would a Consumer get a Bearer token from if not via OAuth?

This would assume that a Consumer already has prior knowledge of a Thing, which is unlikely in a plug-and-play interoperability situation.

benfrancis:
Also, what happens if the token expires?

benfrancis:
@benfrancis wrote:
>  How does the Consumer acquire a token?

I've recently been implementing this on the Consumer side and was able to solve this part by simply prompting the user for a token in the user interface.

That doesn't answer the second part of what to do if and when the token expires. If there is a user interface then you could just prompt the user for a new token the next time they access the user interface (though the Thing would be inaccessible until they did, which would prevent any automations from working).

My understanding is that best practice with token based authentication is to make tokens expire very quickly (e.g. every 10 minutes) and require clients to request a new token via refresh token on regular intervals. In that situation it would be impractical to prompt the user every time.

@mlagally wrote:
> There are significant implementation efforts required to address the code and client flows for OAuth.

Putting the above issue aside for a moment (since not all Bearer tokens will expire quickly, or at all), I do agree with this. If the only options are the `BasicSecurityScheme` (which isn't particularly secure) and the `OAuth2SecurityScheme` (which has a significant implementation burden) then it may limit adoption of the profile, particularly on the Producer side where Things are likely to have more constraints.

> Bearer Tokens are the underlying mechanism - consider lowering the bar on that authentication requirement.

If we assume that Consumers are generally less constrained than Things, one option might be to make the `BearerSecurityScheme` mandatory for Consumers as well, so that Things have middle ground option between `BasicSecurityScheme` and `OAuth2SecurityScheme` (but only have to implement one of them).

Alternatively, if that's too much of an implementation burden for Consumers and if the Bearer scheme really is a subset of the OAuth2 scheme, could it be possible to only make the `BearerSecurityScheme` scheme mandatory for Consumers, but suggest they treat the `OAuth2SecurityScheme` scheme as the Bearer scheme if they don't support the whole of OAuth2? That would be messier, but could potentially guarantee at least a baseline of interoperability. I don't know how well it would work in practice.

benfrancis:
[...]

PR #433, opened 2025-07-22
Duplicate conflicting assertion - closes #367

### LINKED ISSUE #367: PR #364 introduced a duplicate conflicting assertion
Please see my feedback in https://github.com/w3c/wot-profile/pull/364#pullrequestreview-1309349428

@lu-zero @mlagally

sebastiankb:
closes #367 

covers also @egekorkan [comment](https://github.com/w3c/wot-profile/issues/367#issuecomment-3052818788)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/pull/433.html" title="Last updated on Jul 29, 2025, 11:25 AM UTC (103f6b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/433/637100f...103f6b7.html" title="Last updated on Jul 29, 2025, 11:25 AM UTC (103f6b7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -552,7 +552,7 @@
                   <code>OAuth2SecurityScheme</code> with the <code>code</code> or <code>client</code> flow
               </li>
               <li>
-                   <code>ComboSecurityScheme</code> with the <code>oneOf</code> member
+                   <code>ComboSecurityScheme</code> with the <code>oneOf</code> member containing at least one of the schemes above
               </li>
             </ul>
         </div>


lu-zero (approved):

benfrancis:
Unless anything changes with #313 and #347 within the next week, it is currently looking like `BearerSecurityScheme` (and `AutoSecurityScheme`) will not be added to the list of required security schemes for Profiles in 1.0.

If this is the case we need to ask whether it really makes sense to require support for the `ComboSecurityScheme` in 1.0.

I can't think of a sensible use case for combining `NoSecurityScheme` + `BasicSecurityScheme` or `NoSecurityScheme` + `Oauth2SecurityScheme`. So I guess the question is whether it's worth requiring `ComboSecurityScheme` so that a Thing can implement both `BasicSecurityScheme` and `OAuth2SecurityScheme`?

The only other use case I can think of is supporting `OAuth2SecurityScheme` with both the `code` and `client` flows, though I've not yet come across that in practice.

In the meantime it would be useful to split the `ComboSecurityScheme` addition out of this PR because we definitely want to land the rest of the changes.

relu91 (approved):
I think it is ok, more complex use cases can be addressed in the future. I can't join tomorrow, but you can take this as a green light from my side.

PR #436, opened 2025-07-23
Clarify behaviour when an HTTP request or response does not have a body - closes #360

### LINKED ISSUE #360: Clarify the behavior when a response does not have body
We mention in https://w3c.github.io/wot-profile/#http-basic-profile-protocol-binding-invokeaction and other places:

```
Content-Type header set to application/json
A body with an input to the action, if any, serialized in JSON
```

Since some http implementation unhelpfully may pass the empty string to a json deserialiser, might be useful to suggest to either explicitly pass "{}" or not set the Content-Type if there isn't a body.

benfrancis:
I had already added assertions to this effect to the events protocol binding in the HTTP Webhook Profile, but I've reviewed the rest of the text for all other uses of `Content-type` and made the following changes:
- Added an assertion not to include a `Content-type` header in an `invokeaction` request when there is no input
- Added an assertion not to include a `Content-type` header in an `invokeaction` response when there is no output
- Added notes to clarify what is meant by "empty body" for:
  - `invokeaction` request & response in the HTTP Basic Profile
  - `subscribeevent` in the HTTP Webhook Profile
  - `subscribeallevents` in the HTTP Webhook Profile

I don't think a clarification is needed for property related operations because when reading and writing properties there should *always* be a value, even if that value is `null`, so the `Content-type` header should always be set in a `writeproperty` request or a `readproperty` response (as is currently specified). Even for meta operations there would at least be an empty object or array.

I don't think clarification is needed for the HTTP SSE Profile because the `Content-type` is always `text/event-stream` regardless of whether an event has a data payload.

Let me know what you think.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-profile/pull/436.html" title="Last updated on Jul 29, 2025, 10:21 AM UTC (b90abc7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-profile/436/b0a9374...benfrancis:b90abc7.html" title="Last updated on Jul 29, 2025, 10:21 AM UTC (b90abc7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1214,7 +1214,7 @@
           </p>
           <p class="note">
             Valid JSON values such as <code>null</code>, <code>""</code> and <code>{}</code> could be a valid 
-            action input and are not be considered to be an "empty" body.
+            action input and are not considered to be an "empty" body.
           </p>
           <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-6">
             <p>
@@ -1416,7 +1416,7 @@
           </pre>
           <p class="note">
             Valid JSON values such as <code>null</code>, <code>""</code> and <code>{}</code> could be a valid 
-            action output and are not be considered to be an "empty" body.
+            action output and are not considered to be an "empty" body.
           </p>
           <h6 id="async-action-response">Asynchronous Action Response</h6>
           <div class="rfc2119-assertion" id="http-basic-profile-protocol-binding-invokeaction-21">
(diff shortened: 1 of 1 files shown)


egekorkan (approved):

benfrancis:
Thanks @TallTed, I have applied your suggested changes and made the same changes elsewhere in the document to match.

TallTed (approved):

benfrancis:
@relu91 wrote:
> Do we want to explicitly set the returned code for empty content-type to [204](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.5)?

Good idea, I've re-written the assertions for the synchronous `invokeaction` response and split into two cases:
1. output
2. no output

@relu91 Do you approve of this PR with these changes?

relu91 (approved):
yup, thanks

benfrancis:
Profile 2025-07-30 Decided to merge.

PR #484, opened 2025-10-28
Remove link to missing atrisk CSS file

benfrancis:
There is a link to atrisk.css which is not included in the publication folder, so results in a 404 error, which is apparently blocking publication.

I've removed the link to [atrisk.css](https://github.com/w3c/wot-profile/blob/main/testing/atrisk.css) since all of the lines of that file are commented out anyway because the assertion numbers are obsolete since we re-worked the specification.

This is a one line change in index.html to remove a broken link that is not currently needed. The rest of the diff is auto-generated by Respec and tidy.


benfrancis:
r=me to unblock publication.

PR #490, opened 2025-12-10
Clarify how the Profiling mechanism will work - closes #488

### LINKED ISSUE #488: Requirement: Clarify how the profiling mechanism will work
It was noted in #486 that currently the use cases & requirements document talks a lot about about the requirements for the registry definition and registry entries, but not a lot about how the profiling mechanism itself will work.

We should define additional requirements which outline how the profiling mechanism is expected to work. (It really isn't that complicated, it just uses the `profile` member of a TD to list the URIs of the Profiles to which the Thing conforms).

benfrancis:
This PR adds additional detail to the Use Cases & Requirements document about how the profiling mechanism is expected to work.

Closes #488.

### DIFF
--- a/requirements/index.html
+++ b/requirements/index.html
@@ -104,6 +104,9 @@
       <ol>
         <li>The specification shall define a mechanism by which the author of a WoT Thing Description can denote that
           the Thing it describes conforms to a particular profile.</li>
+        <li>The profiling mechanism shall use the <code>profile</code> member of a Thing Description
+          [[thing-description-2.0]] to enumerate the URIs identifying the Profiles to which the Thing it describes
+          conforms.</li>
       </ol>
     </section>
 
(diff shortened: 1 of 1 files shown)


relu91 (approved):

requirements/index.html
```diff
@@ -165,7 +168,7 @@ <h3>Profile Registry Entries</h3>
           document.</li>
         <li>Profiles shall not include any assertions which would cause a Thing Description of a conformant Web Thing
           to be non-conformant with the Thing Description 2.0 specification.</li>
-        <li>Profiles registered in the Profile Registry should avoid including assertions which conflict with 
+        <li>Profiles registered in the Profile Registry should avoid including assertions which conflict with
```
TallTed, line 171:
```suggestion
        <li>Profiles registered in the Profile Registry should not include assertions which conflict with
```

benfrancis: Please see comment in https://github.com/w3c/wot-profile/pull/491/files#r2626194740

TallTed (approved):

